PR #19, opened 2016-10-11
Add metadata block. Fix button styles.

alexander-yevsyukov:


alexander-yevsyukov:
The script also moved from `/js` into `/gmscripts` directory.

PR #42, opened 2018-05-09
Add a placeholder for JEEConf'18 landing pages.

armiol:
In order to improve visitor tracking, we are creating a https://spine.io/jee18 page.

Another page created in this PR is https://spine.io/alexander placeholder to track those who used the QR code from the Alexander's business card.

This PR creates empty pages which immediately redirect a user to the homepage, but send the analytics report beforehand.

### DIFF
--- a/alexander/index.html
+++ b/alexander/index.html
@@ -1,0 +1,11 @@
+---
+layout: default
+group: 'navigation'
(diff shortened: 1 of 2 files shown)


armiol:
@alexander-yevsyukov PTAL.

PR #62, opened 2018-10-17
Fix navigation

JuliaEvseeva:
This PR brings a new header navigation that will be adaptive for the desktop and mobile devices. Thanks to this, changes in the navigation items will be only in one place. Additional `sticky-nav` and `mobile-nav` were removed.

Also, the footer for the mobile devices was fixed.
Fixed `pre code` tag. Now it resizes with layout columns.


_sass/modules/_nav.scss
```diff
@@ -1,105 +1,289 @@
...
-    display: none;
-    font-size: 13px;
-    letter-spacing: 1px;
-    text-align: right;
-    position: relative;
-    z-index: 9012;
+//Navigation
+.navigation-opened {
+  @media(max-width: $tablet){
+    overflow: hidden;
```
andreySadomovskiy: Improve behavior for open/close  states. 
We have wrong position of  `close` icon in `navigation-opened` state.  Because of hidden scroll and jumping content by same reason. Add please `17px` compensation.    

image

js/common.js
```diff
@@ -1,3 +1,25 @@
...
+
+//Mobile navigation — doc-list toggle
+$('.doc-list-toggle').on('click', function(){
+    $('.doc-list-inside').toggleClass('active');
+    $(this).toggleClass('active');
+});
+
+//Mobile footer navitation
+$(document).ready(function() {
+    var $window = $(window);
```
andreySadomovskiy: Not used variable `var $window = $(window);`. Remove it, please.

js/common.js
```diff
@@ -1,3 +1,25 @@
...
+    $('.toggle').on('click',function(){
```
andreySadomovskiy: Move this part of code to `$(document).ready(function() {}`  upper.  We already have this event.

js/common.js
```diff
@@ -268,3 +169,26 @@ function FixToc() {
...
+            header.addClass("pinned"); //When navigation below hero section
+            header.removeClass("unpinned");
+        }
+        else {
+            header.removeClass("pinned");
+            header.addClass("unpinned");
+        }
+
+        //Return classes to the initial state when the navigation at the top of the page
+        if (window.pageYOffset < 64) {
```
andreySadomovskiy: Could you take this value — 64  by JS from container  `.innerHeight();` ? To make it more flexible for styles changes. 

For this `tocNav.tocify({selectors:"h2, h3, h4", showAndHide: false, scrollTo: 64, extendPage: false});` place as well.

_sass/modules/_nav.scss
JuliaEvseeva: It is not necessary for mobile devices. 
We can add this only for the desktop devices for the Windows.

andreySadomovskiy: Got it, thanks.  Let's leave it on this stage.

PR #71, opened 2018-11-09
Update dark theme implementation

JuliaEvseeva:
This PR brings dark theme for the code that loads immediately without a jump to the white theme.


js/common.js
```diff
@@ -28,7 +28,7 @@ $("i.fa-github-alt").parent().removeClass("external");
 
 // Prettyprint
 $('pre').addClass("prettyprint");
-$.getScript("https://cdn.rawgit.com/google/code-prettify/master/loader/run_prettify.js", function(){
+$.getScript("/js/run_prettify.js", function(){
```
andreySadomovskiy: Merge and check if this URL will be available on server

PR #94, opened 2018-11-26
Update feature section and add style for the note text

JuliaEvseeva:
This PR brings:
- updated feature section — now feature cards have their height and the same distance between cards;
- updated icons from fontawesome.com site;
- added style for the note block.

### DIFF
--- a/_sass/modules/_feature-cards.scss
+++ b/_sass/modules/_feature-cards.scss
@@ -1,5 +1,5 @@
 .features-section {
-  background-color: $light-gray;
+  background-color: $light-gray-blue;
   padding: 72px 0 80px;
 
   @media(max-width: $phone-xlarge) {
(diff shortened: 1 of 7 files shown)


JuliaEvseeva:
@andreySadomovskiy, PTAL

_sass/modules/_feature-cards.scss
```diff
@@ -56,37 +41,9 @@
       height: $icon-size--m;
       margin: 20px 24px;
 
-      @media (max-width: $phone-medium) {
+      @media (max-width: $desktop-small) {
         margin: 16px 16px;
```
andreySadomovskiy: You can use shorter variant  `margin: 16px;` if you wish.

JuliaEvseeva: Done.

PR #98, opened 2018-11-26
Why Spine

alexander-yevsyukov:
This PR updates the home page with reasons to love Spine.

It also simplifies the text on immutability.


alexander-yevsyukov:
@JuliaEvseeva, PTAL.

PR #99, opened 2018-11-26
Improve about

JuliaEvseeva:
This PR brings:
- updated styles for the headline text on the `about` page.
- links to client libraries.


JuliaEvseeva:
@alexander-yevsyukov, PTAL

PR #100, opened 2018-11-26
Add client lib link

alexander-yevsyukov:
This PR adds a client-side Java API reference link, and arranges the links in language-specific boxes.

### DIFF
--- a/_sass/pages/_docs.scss
+++ b/_sass/pages/_docs.scss
@@ -9,6 +9,10 @@
     }
   }
 
+  .doc-lead {
+    margin-top: 32px;
+  }
+
   .info-section {
     margin: 0 10px 0;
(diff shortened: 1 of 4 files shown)


alexander-yevsyukov:
@JuliaEvseeva, PTAL.

You may notice empty `<li>` I've added because boxes with only one link were smaller than the Java box with two links. It would be better to allow a minimum size for such a box. Can you do that, please?

alexander-yevsyukov:
@andreySadomovskiy, could you PTAL, while Julia is away?

andreySadomovskiy:
@alexander-yevsyukov , LGTM.  Also we need general adjustment for content layout and styles to simplify content management #101, will discuss this with Julia later.

PR #103, opened 2018-12-04
Fix cookie notification button

JuliaEvseeva:
This PR brings updated sizes for the cookie notification “Agree” button.
Also, `translate3D` was changed on `translate` to remove blurred pixel in the text.


JuliaEvseeva:
@andreySadomovskiy, PTAL

PR #104, opened 2018-12-05
Update docs pages

JuliaEvseeva:
This PR brings:
- updated styles for the docs pages;
- fixed `ol` `ul` list styles;
- fixed `toc` height bug if the number of items doesn't fit on the page;
- added fontawesome-pro version;
- updated icons for the “Clear API” and “Choice of Storage and Deployment Platforms” feature cards.


JuliaEvseeva:
@andreySadomovskiy, PTAL

Issue #141, opened 2019-04-05
Update Java Quick Start guide


alexander-yevsyukov:
It's needed to import both `time.proto` and `time_options.proto`. The latter is needed for being able to do `(when).in = FUTURE`. It's the recent change.

- [ ] Explain validation options and link to Validation User Guide
Also we need to mention something about the validation options that we add to field declarations. We do not need to give all the details, but explain what we're basically doing here. And then give the link to a user guide on the Validation Library.

- [ ] Use `vBuilder()` API instead of `XxxVBuilder.newBuilder()`
   This is a recent change that we'd like to promote. The current code is still valid, but it's easier to explain `vBuilder()` vs `newBuilder()` thing.
   It's basically this. `vBuilder()` creates a Validating Builder, which checks the values against the validation options when a message is going to be build. `newBuilder()` creates a standard `Builder` natively provided by Protobuf. It is safer to always use `vBuilder()`, and it does not make much sense to specify validating options, if `vBuilder()` is not used. For example,  commands are always validated upon arrival to the server side. But it's too late. It's better to create a valid command on the client side.

- [ ] Fix the code alignment in `requestFactory.command().create()` call
   It should be either aligned by dots, or be on the same line. 

- [ ] Make the code indentation in the code fragments consistent. At the moment `AssignDueDate` has no indentation at all, other messages have their content indented with just a single space character. The indentation should be two spaces.

- [ ] Numeration of the points in "What’s Next" section is broken. We have two ones there.

meizabel:
@armiol I'm assigning it to @JuliaEvseeva to items numbering fix. At  the moment it's spine.io css limitations.

armiol:
@meizabel it's not a CSS limitation.

armiol:
@meizabel please see [this branch](https://github.com/SpineEventEngine/SpineEventEngine.github.io/tree/fix-whats-new-item-numbering) (there's just a single commit on top of `master`) and let me know if this works for you.

JuliaEvseeva:
@armiol, unfortunately this is a markdown limitation. We cannot put `<p>` tag inside the markdown list. To fix this we should use HTML `<ol><li></li></ol>` tags.

meizabel:
@armiol 
LGTM 
Thanks for your support!

meizabel:
Closing this issue. 
All requited changes have been implemented and merged into a master.

PR #142, opened 2019-04-08
Fixes for issue #141:

meizabel:
In Java Quick Start guide (quickstart/java.md):
1. Fixed the code alignment in requestFactory.command().create().
2. Added a note on vBuilder and fixed vBuilder() vs newBuilder() naming.
3. Added Validation User Guide to source control and created a link to this guide from Java Quick Start Guide and spine.io/about page, Built-in Validation feature description.

### DIFF
--- a/docs/quickstart/java.md
+++ b/docs/quickstart/java.md
@@ -119,7 +119,7 @@ You do not have to make any additional steps to use it in your domain.</p>
  }
   ```
 <p class="note">`vBuilder()` creates a Validating Builder, which checks the values against the validation options when a message is going to be build. The `newBuilder()` creates a standard Builder natively provided by Protobuf. It is safer to always use `vBuilder()`, and it does not make much sense to specify validating options, if `vBuilder()` is not used.
-For example, commands are always validated upon arrival to the server side. We recommend creating a valid command on the client side as it can be too late if the validation is performed upon arrival on the server side.</p> 
+For example, commands are always validated upon arrival to the server side. We recommend creating a valid command on the client side as it can be too late if the validation is performed upon arrival on the server side. For more details, please refer <a href="docs/guides/validation-user-guide.html">Validation User Guide</a>.</p> 
 5. Apply the emitted event:
   ```java
  @Apply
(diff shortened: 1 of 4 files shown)


docs/quickstart/java.md
```diff
@@ -113,12 +112,14 @@ Keep experimenting with your model. To do so:
...
  DueDateAssigned handle(AssignDueDate command) {
      return DueDateAssignedVBuilder
-             .newBuilder()
+             .vBuilder()
              .setTaskId(command.getTaskId())
              .setDueDate(command.getDueDate())
              .build();
  }
   ```
+<p class="note">`vBuilder()` creates a Validating Builder, which checks the values against the validation options when a message is going to be build. The `newBuilder()` creates a standard Builder natively provided by Protobuf. It is safer to always use `vBuilder()`, and it does not make much sense to specify validating options, if `vBuilder()` is not used.
```
alexander-yevsyukov: When we replace `newBuilder()` with `vBuilder()`, then the standard Protobuf API is not visible to the user. Therefore the reference looks strange. The user needs to know Protobuf API sufficiently, and in most cases that would be not the case.

docs/quickstart/java.md
```diff
@@ -75,9 +75,8 @@ Keep experimenting with your model. To do so:
      spine.time.LocalDate due_date = 2 [(valid) = true, (required) = true, (when).in = FUTURE];
  }
   ```
-<p class="note">Remember to import `LocalDate` via `import "spine/time/time.proto";`. This type is provided by
-                  the [Spine Time](https://github.com/SpineEventEngine/time) library. You do not have to make any
-                  additional steps to use it in your domain.</p>
+<p class="note">Remember to import `LocalDate` using `import "spine/time/time.proto";` and `import "spine/time/time.options.proto";`. The latter is needed for being able to do `(when).in = FUTURE`. This type is provided by the [Spine Time](https://github.com/SpineEventEngine/time) library.
```
alexander-yevsyukov: The file name is `time_options.proto`

docs/quickstart/java.md
meizabel: Maybe this note should be an add-on for Validation User Guide rather than Java Quick Start Guide? Or should I better reference to a Validation Guide for more details?

Issue #137, opened 2019-04-05
Add “Event Reactor” to Messaging concepts


alexander-yevsyukov:
Suggested text:

An Event Reactor is an object which usually produces one or more events in response to an incoming event. Unlike an Event Subscriber, which always consumes events, a reacting object generates events in response to changes in the domain.

meizabel:
Closing the issue as its contents has already been merged into a master branch.

PR #145, opened 2019-04-08
Fixing issue #137

meizabel:
- Added Event Reactor concept contents to version control
- Modified doc-side-concepts-nav.html file to reflect new Concepts section structure

### DIFF
--- a/docs/concepts/event-reactor.md
+++ b/docs/concepts/event-reactor.md
@@ -8,4 +8,7 @@ type: markdown
 ---
 <h2 class="top">Event Reactor</h2> 
 Event Reactor is an object which usually produces one or more events in response to an incoming event. 
-Unlike <a href="https://spine.io/docs/concepts/event-subsriber.html">Event Subscriber</a>, which always consumes events, a reacting object generates events in response to changes in the domain.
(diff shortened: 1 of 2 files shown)


docs/concepts/event-reactor.md
```diff
@@ -0,0 +1,11 @@
...
+title: Event Reactor
+headline: Event Reactor
+bodyclass: docs
+layout: docs
+sidenav: doc-side-concepts-nav.html
+type: markdown
+---
+<h2 class="top">Event Reactor</h2> 
+Event Reactor is an object which usually produces one or more events in response to an incoming event. 
+Unlike <a href="https://spine.io/docs/concepts/event-subsriber.html">Event Subscriber</a>, which always consumes events, a reacting object generates events in response to changes in the domain.
```
alexander-yevsyukov: Not “always”. In some cases, a Reactor may ignore the event, returning `Nothing`. It usually happens when a tuple is declared in the return type of a method, and `Nothing` is one of the possible options.

alexander-yevsyukov: For example, there could be a return type declared like `EitherOf2<TaskReAssigned, Nothing>`.

Issue #150, opened 2019-04-09
Add a link to the guide from Rejection concept to Creating Rejection Messages User Guide


meizabel:
Refer readers to [Creating Rejection Messages Guide](docs/guides/creating-rejection-messages.html)

meizabel:
Added a reference to Defining Rejections Guide.

PR #151, opened 2019-04-09
Adding a reference to Creating Rejection Messages Guide to address issue #150.

meizabel:
In rejection.md file added a reference to Creating Rejection Messages Guide to address issue #150 and improve the documentation user experience.

### DIFF
--- a/docs/concepts/rejection.md
+++ b/docs/concepts/rejection.md
@@ -17,4 +17,4 @@ Consider the following examples of rejections:
 * `OrderCannotBeEmpty`, 
 * `InsufficientFunds`.
 
-For detailed instructions on creating rejection messages, refer to [Defining Rejection Messages Guide](/docs/guides/creating-rejection-messages.html).
+For detailed instructions on creating rejection messages, refer to [Defining Rejections Guide](/docs/guides/creating-rejection-messages.html).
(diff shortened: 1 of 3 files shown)


docs/concepts/rejection.md
```diff
@@ -15,4 +15,6 @@ In Spine, rejections are defined as Protobuf messages. If an event is a fact of
 Consider the following examples of rejections: 
 * `CreditCardValidationDeclined`, 
 * `OrderCannotBeEmpty`, 
-* `InsufficientFunds`.
\ No newline at end of file
+* `InsufficientFunds`.
+
+For detailed instructions on creating rejection messages, refer to [Creating Rejection Messages Guide](/docs/guides/creating-rejection-messages.html).
```
alexander-yevsyukov: I think we should name the document differently. There are three kinds of messages the user defines in a Bounded Context: events, commands, and rejections (as special kind of, so to say, “negative” events).

Technically, these things are messages in the sense that they are objects that are sent between the computation nodes, and in the sense that they all implement the `Message` interface.

But when we speak about them, we don't really need to mention the word “message” because it is assumed that an event, or a command _is_ a message.

So, I'd name the guide “Defining Rejections” along with “Defining Events” and “Defining Commands”.

meizabel: Ok, I will rename the guide to "Defining Rejections". 

Two more questions: 
1. Can we come out with Defining Events and Defining Commands guides texts then? 
2. Also, can I add the explanation of messages defined within a Bounded Context to a Messaging section in Concepts?

Issue #128, opened 2019-04-04
Add a new Spine.io product description to the website


meizabel:
Current proposal to be discussed: "Apply reactive Domain-Driven Design quicker, with less code transforming EventStorming into a domain model code." 

So far looks rather overloaded and needs to be simplified.

meizabel:
Also, in footer.html file we have the following entry to think about: 
"CQRS/ES framework for modern cloud applications".

meizabel:
Closing this issue as the description has been updated recently after discussion with Spine developers.

PR #180, opened 2019-04-18
Revising product description within the framework of #128

meizabel:


meizabel:
Deleting this one for the moment as we have come to a different solution on product description.

PR #181, opened 2019-04-19
Add a possibility to open the `Architecture Diagram` in full screen mode

JuliaEvseeva:
In this PR a possibility to open a `Spine Architecture Diagram` in full-screen mode was added.
Also, a new `full-screen` layout was created.


JuliaEvseeva:
@alexander-yevsyukov, @andreySadomovskiy, PTAL

css/style.scss
```diff
@@ -13,6 +13,8 @@
 @import "base/_text";
 @import "base/_override";
 @import "base/_markdown";
+@import "base/_modals";
```
andreySadomovskiy: Do we need this file now?

Issue #154, opened 2019-04-10
Add Integration Event concept contents


meizabel:

meizabel:
@armiol 
Can we please come out with some ideas for this concept? 
I would really appreciate it.

meizabel:
Closing issue. Now it is live on spine.io website, Concepts section.

PR #192, opened 2019-05-13
Add `IntegrationEvent` concept

armiol:
This changeset updates the "Integration Event" page with some content.

Also, an order of the items in the "Concepts" navigation bar was changed to put "Integration Event" lower than the more important concepts such as "Command Bus".

This PR addresses #154.


armiol:
@meizabel PTAL.

PR #199, opened 2019-07-24
Hide a `toc` functionality if a page has less than 3 sub-headers

### LINKED ISSUE #123: Hide a `toc` functionality if a page has less then 3 sub-headers
The goal of this task is to not show `toc` links at the right of the page if there is less then 3 headers.

image

JuliaEvseeva:
This PR closes #123.

In this PR I have added a new function that hides a `toc` functionality if a page has less than 3 sub-headers.

As part of the PR, I have also fixed a `doc-side` navigation and used a `Jekyll Liquid` to make sure that the current state matches the selected item in the `doc-side` navigation.


JuliaEvseeva:
@andreySadomovskiy, PTAL.

PR #200, opened 2019-07-25
Use angle brackets for the generic parameters in the code example

JuliaEvseeva:
In this PR I have used angle brackets for the generic parameters in the example code on the `Event Subscriber` page.


JuliaEvseeva:
@armiol, PTAL.

JuliaEvseeva:
@alexander-yevsyukov, PTAL.

PR #213, opened 2019-08-12
Fix `body` scroll jumping

### LINKED ISSUE #210: Fix `body` scroll on Windows
When switching pages in the Documents section, content jumps while the page is loading. The problem is that the scroll on Windows does not appear immediately. It's hidden and then appears.

JuliaEvseeva:
This PR fixes #210.

This PR fixes a bug with appearing a `body` scroll on Windows.
It has happened because the custom js scripts were loaded before the `common.js`.
In this PR I have changed the order of the custom js scripts including and place them after the `common.js`.


JuliaEvseeva:
@andreySadomovskiy, PTAL.

PR #215, opened 2019-08-21
Fix back to blog button link

andreySadomovskiy:
Back to the blog button was fixed in this PR.


andreySadomovskiy:
@JuliaEvseeva , PTAL

andreySadomovskiy:
@JuliaEvseeva , now back button scrolls to the needed article. But we also need to implement `scrollTo` with `offset` logic.  Offset it is header height value.

JuliaEvseeva:
The `cookie` notification brokes now because of hashes in the post `ID`.

PR #219, opened 2019-09-04
Fix scroll behavior

JuliaEvseeva:
In this PR I have updated scroll behavior to the anchor point on the `docs` and `faq` pages.

I have separated functions that are used only for the FAQ page. And also, I have adjusted the offset value for the `docs` pages.

All mixins were extracted to the `base/_mixins.scss` file.


JuliaEvseeva:
@andreySadomovskiy, PTAL.

PR #220, opened 2019-09-05
Adjust `post` page paddings

JuliaEvseeva:
In this PR I have adjusted paddings on the `post` page to add more space between footer and content.


JuliaEvseeva:
@andreySadomovskiy, PTAL.

PR #221, opened 2019-09-05
Cleanup fonts import

JuliaEvseeva:
In this PR I have removed duplicated font import in the `base.html` file.
Also, I have removed the import of missing font `roboto-mono-v5-latin-100`. The font isn't used in the project yet.


JuliaEvseeva:
@andreySadomovskiy, PTAL.

PR #222, opened 2019-09-09
Add theme color switch

JuliaEvseeva:
In this PR I have added a `color-selector` that appears at the right of the window and changes code theme color. The `color-selector` is visible only if a page has a `pre` element. On the `promo` page the `color-selector` will appear on scroll if it overlaps the `hero-section`.

The `theme-color` variable will be kept in cookies. For that reason `/libs/js-cookie/js.cookie.js` is added.

Also, bootstrap `tooltips` were added and styled.


JuliaEvseeva:
@andreySadomovskiy, PTAL.

js/common.js
```diff
@@ -1,24 +1,50 @@
...
 
+/** Code color variables */
+const $colorSelector = $('.color-selector');
+const $colorSelectorLinks = $('.color-selector .color-link');
+const $selectorDark = $('.color-link.dark');
+const $selectorLight = $('.color-link.light');
+const colorDark = 'dark';
+const colorLight = 'light';
+const cookieColorName = 'themeColor';
+const baseStylesUrl = '/libs/prettify/skins/';
```
andreySadomovskiy: I would say it is not  `baseStylesUrl`.  It is code wrapper skins url or something similar.

PR #226, opened 2019-10-15
Update `nokogiri` and `rubyzip` modules versions

andreySadomovskiy:
`nokogiri` and `rubyzip` modules versions were updated in this PR to fix security vulnerability issues.


andreySadomovskiy:
@JuliaEvseeva , PTAL

PR #227, opened 2019-12-11
Fix `Go-to-top` button

JuliaEvseeva:
This PR brings a fixed `go-to-top` button. The problem was that the button wasn't shown on the page.


JuliaEvseeva:
@andreySadomovskiy, PTAL.

PR #231, opened 2019-12-27
Fix `Toc` hash generation

JuliaEvseeva:
This PR brings a custom hash generation function that removes all punctuation symbols from the hash.

It fixes a bug when the generated hash has some punctuation symbol and it doesn't match with the title id. For example `improved-areas:` in the hash and `improved-areas` in the title.


JuliaEvseeva:
@andreySadomovskiy, PTAL.

PR #233, opened 2020-01-21
Remake doc navigation

JuliaEvseeva:
The PR brings a new structure for the document navigation menu.
The top navigation and side navigation were combined into one side navigation.

Also, as a part of the PR, `sticky-elements` functions were moved to the separate file.

General `.md` file cleanup should be done with #234 issue.


JuliaEvseeva:
@alexander-yevsyukov, PTAL.

PR #235, opened 2020-01-21
Improving navigation

alexander-yevsyukov:
This PR:
  * Adds Client Libraries section.
  * Adds Introduction section and moves Project Structure and Naming Convention pages there.
  * Removes C++ from the text where found. We're not going to support this language.


_data/navigation/doc_side_nav.yml
```diff
@@ -119,7 +113,27 @@ doc_side_nav:
...
+        sub_sub_nav_id: lib-java
+        sub_sub_nav:
+          - page: API Reference
+            url: https://spine.io/core-java/javadoc/client/index.html
+      - page: JavaScript
+        url: '#lib-js'
+        sub_sub_nav_id: lib-js
+        sub_sub_nav:
+          - page: API Reference
+            url: https://spine.io/docs/reference/javascript/index.html
```
JuliaEvseeva: Now this link does not work. Are we planning to add it in the future?
The same for the `Dart API Reference`.

alexander-yevsyukov: They don't. There are going to be real things anytime soon.

PR #236, opened 2020-01-22
Simplify the document side navigation

JuliaEvseeva:
This PR brings simplified attribute names for side navigation.
Also, the `REAMDE.md` file was updated.

The new structure for the three-level navigation list:
```
doc_side_nav:
  - page: Messaging
    children:
      - page: Architecture Overview
        url: /docs/architecture-overview.html
      - page: Event
        children:
          - page: New page on the 3rd level
            url: /docs/3rd-level-file.html
```

The `id` and `href` that are needed to create a collapsible navigation item now will be generated automatically based on a page name in the `_includes/doc-side-nav.html` file.

Result:

image

### DIFF
--- a/_data/navigation/README.md
+++ b/_data/navigation/README.md
@@ -49,7 +49,7 @@ tree by clicking on it.
 #### Example for the three-level navigation list
 
 ```
-concepts:
+doc_side_nav:
   - page: Messaging
     children:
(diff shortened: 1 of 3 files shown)


JuliaEvseeva:
@alexander-yevsyukov, PTAL.

PR #239, opened 2020-01-24
Remove top document navigation that is no longer in use

### LINKED ISSUE #238: Remove unused `doc-nav.html` template
The goal of this task is to remove the `doc-nav.html` template, which is no longer in use. 
And remove all styles and JS that are associated with it.

JuliaEvseeva:
This PR closes #238.

In this PR I have removed the top document navigation that is no longer in use.
Removed all styles and JS that were associated with it.

Also, I have renamed the `_doc-nav.scss` file to `_doc-side-nav.scss`, because there are only styles for side navigation.


JuliaEvseeva:
@alexander-yevsyukov, PTAL.

PR #241, opened 2020-01-24
Update `blockquote` styles

JuliaEvseeva:
This PR brings updated styles for the `blockquote`.

Also, for the external links, the `target="_blank"` attribute (that opens the external link in a new browser tab) will be added automatically.  It's useful for the markdown links that cannot be extended by the HTML attributes.

The FAQ page layout was cleaned up.


JuliaEvseeva:
@alexander-yevsyukov, PTAL.

JuliaEvseeva:
@alexander-yevsyukov, PTAL.

PR #248, opened 2020-01-31
Update the feature section

JuliaEvseeva:
This PR brings the updated feature section to the `Home` and `About` pages.
The functions were combined from these two pages.


JuliaEvseeva:
@alexander-yevsyukov, PTAL.

PR #255, opened 2020-02-11
Copy the URL to clipboard by clicking on the anchor heading

JuliaEvseeva:
This PR brings new functionality that allows copying the URL to the clipboard by clicking on the anchor link near the heading. After clicking the snackbar notification will be shown at the left bottom corner.


JuliaEvseeva:
@alexander-yevsyukov, PTAL.

PR #256, opened 2020-02-12
Fix UI issues

JuliaEvseeva:
This PR brings fixes for small UI issues:

- `C++` was replaced with the `Dart` on the home page in the `Step 2` text.
- The header layout for the architecture diagram in full-screen was fixed.
- Heading styles were defined again to don't depend on `h1` style changes.


JuliaEvseeva:
@alexander-yevsyukov, PTAL.

PR #267, opened 2020-03-30
Update references to generated doc

dmdashenkov:
Since recently, more generated documentation is published to `spine.io`. In this PR we add links to the new docs and update links to the docs which have changed their directory structure.

Also, the link to "API Reference" on the home page now leads to the general "Docs" page, as we've got no special page for API reference overview and I'm not convinced that we need one.


dmdashenkov:
@armiol, @alexander-yevsyukov, PTAL.

PR #269, opened 2020-04-12
Validation doc

dmdashenkov:
In this PR we add the Spine [validation user guide](https://spine.io/docs/guides/validation-user-guide.html).

### DIFF
--- a/docs/guides/validation-user-guide.md
+++ b/docs/guides/validation-user-guide.md
@@ -11,9 +11,8 @@ something more than just a data structure. One of the commodities required for d
 specifics is data validation. Spine extensively uses the Validation library when working with any
 Protobuf messages.
 
-The purpose of the Validation library is giving framework users an easy way to:
- - maintain simple domain invariants;
- - construct simple Anticorruption layers.
+One way or another, the framework users will have to validate entity states. We provide 
+the Validation library so that you don't have to write it by hand.
 
 This guide will walk you though the API of Spine Validation library. All of the validation features
 described here are currently supported in the Java environment. Many are supported in Dart as well.
--- a/docs/guides/validation-user-guide.md
+++ b/docs/guides/validation-user-guide.md
@@ -39,13 +39,50 @@ transfer messages over the wire when needed.
 #### How required fields work
 
 Fields in Protobuf may have either a primitive type or a user-defined type. A user-defined type is
-a `message` or an `enum` and primitive types are numbers, `string`, and `bytes`. Due to limitations
-of the binary format, there is no way to tell if a numeric field is set to `0` or not just not set.
+a `message` or an `enum` and primitive types are numbers, `string`, and `bytes`. If a `message` or
+an `enum` field is not set, the default value is assigned automatically:
+  
+```java
+ZonedDateTime time = getZonedTime();
+ZoneId timeZone = time.getZoneId();
+assert timeZone.equals(ZoneId.getDefaultInstance());
+```
+
+However, due to limitations of the binary format, there is no way to tell if a numeric field is set
+to `0` or not just not set:
+
+```java
+LocalTime time = getTime();
+int hours = time.getHours();
+assert hours == 0;
+```
+
 This means that a numeric field cannot be required, as there is no way to check if it is set. All 
 the other fields can be required. For `message` fields this means that the message must not be
-empty. For `enum` fields, this means that the enum value must have a number other than `0` (since 
-the enum value with number `0` is the default value of the field). For `string` and `bytes` fields
-this means that the sequence must not be empty.
+empty:
+
+```java
+ZonedDateTime time = getZonedTime();
+LocalDateTime dateTime = time.getDateTime();
+assert !dateTime.equals(LocalDateTime.getDefaultInstance());
+```
+ 
+For `enum` fields, this means that the enum value must have a number other than `0` (since 
+the enum value with number `0` is the default value of the field):
+
+```java
+LocalDate date = getDate();
+Month month = date.getMonth();
+assert !month.equals(Month.MONTH_UNDEFINED);
+```
+ 
+For `string` and `bytes` fields this means that the sequence must not be empty:
+
+```java
+PersonName name = getName();
+String givenName = name.getGivenName();
+assert !givenName.isEmpty();
+```
 
 For collection fields (i.e. `repeated` and `map`), a field is considered set if:
   1. The collection is not empty.
--- a/docs/guides/validation-user-guide.md
+++ b/docs/guides/validation-user-guide.md
@@ -30,8 +30,8 @@ MyMessage.newBuilder()
          .vBuild(); // ← Throws ValidationException. 
 ``` 
 
-If validation is not required, you may call `build()` or `buildPartial()` to symbolize that 
-the message is likely to be invalid.
+If the validation is not required, you may call `build()` or `buildPartial()` provided by Protobuf
+Java API.
 
 The message class also gets an extra method — `validate()`. This method does not throw exceptions.
 Instead, it returns a list of `ConstraintViolation`s:
@@ -292,7 +292,7 @@ In this case, the `ProfilePictureChanged.id` field is not required, since it's n
 in the field. The field `ProfilePictureChanged.new_picture` is not required because the convention
 is overridden with an explicit option.
 
-### Recursive message validation
+### Nested message validation
 
 When a message is validated, only the "shallow" constraints are checked by default. This means that
 the message fields can be invalid and the container message is still considered valid.
--- a/docs/guides/validation-user-guide.md
+++ b/docs/guides/validation-user-guide.md
@@ -409,7 +409,3 @@ of such constrains in the whole project is not large, significantly smaller than
 normal constraints. This mechanism is not designed to override validation rules of an entire library
 of Protobuf definitions, merely a small amount of local patches.
 </p>
-
-# Custom constraints
-
-
--- a/docs/guides/validation-user-guide.md
+++ b/docs/guides/validation-user-guide.md
@@ -79,6 +79,8 @@ Other cases call for conditional required fields. In particular, some fields of 
 set alongside other fields. Consider an example of an online store item:
 
 ```proto
+import "spine/options.proto";
+
 message Item {
 
     // ...
@@ -146,6 +148,8 @@ convention, use `(required) = false`.
 Example:
 
 ```proto
+import "spine/options.proto";
+
 message User {
     option (entity).kind = AGGREGATE;
 
@@ -161,6 +165,8 @@ has nothing to do with this convention, only the field __order__. Thus, `User.na
 For the next example, consider `user_events.proto`:
 
 ```proto
+import "spine/options.proto";
+
 message ProfilePictureChanged {
 
     Url new_picture = 1 [(required) = false];
@@ -180,6 +186,8 @@ the message fields can be invalid and the container message is still considered 
 In order to enable message field checks, use `(validate)` option:
 
 ```proto
+import "spine/options.proto";
+
 message User {
     
     PersonName name = 2 [(validate) = true];
@@ -198,6 +206,8 @@ If a specific error message is required for an invalid field, the `(if_invalid)`
 used:
 
 ```proto
+import "spine/options.proto";
+
 message User {
     
     PersonName name = 2 [(validate) = true,
@@ -221,6 +231,8 @@ By default, the bounds are __inclusive__. Use the `exclusive` property to make a
 Example:
 
 ```proto
+import "spine/options.proto";
+
 message Distance {
 
     uint64 meters = 1;
@@ -238,6 +250,8 @@ boundary and a square bracket (`[` or `]`) — an inclusive boundary.
 Example:
 
 ```proto
+import "spine/options.proto";
+
 message LocalTime {
     
     int32 hours = 1 [(range) = "[0..23]"];
@@ -277,6 +291,8 @@ expression to match the field values. Also, some common pattern modifiers are av
 Example:
 
 ```proto
+import "spine/options.proto";
+
 message HyperReference {
     string url = 1 [(pattern) = { 
             regex: "https?://.+\\..+" 
@@ -333,6 +349,8 @@ a `map` field to only contain non-duplicating elements (values in case of `map`s
 Example:
 
 ```proto
+import "spine/options.proto";
+
 message User {
 
     repeated EmailAddress recovery_emails = 42 [(distict) = true];
@@ -352,6 +370,8 @@ from a non-default value to anything else will cause a violation.
 Example:
 
 ```proto
+import "spine/options.proto";
+
 message Order {
     option (entity).kind = AGGREGATE;
 
@@ -384,6 +404,8 @@ constraints defined outside the message.
 To declare an external constraint, use the `(constraint_for)` option:
 
 ```proto
+import "spine/options.proto";
+
 message UserPictureConstraint {
     option (constraint_for) = "org.example.user.User.profile_picture";
 
--- a/docs/guides/validation-user-guide.md
+++ b/docs/guides/validation-user-guide.md
@@ -19,6 +19,48 @@ This guide will walk you though the API of Spine Validation library. All of the 
 described here are currently supported in the Java environment. Many are supported in Dart as well.
 For more info, see the description of individual constraints.
 
+## Java validation API
+
+For Java, we generate additional code to the Protobuf message classes. In particular, the message
+builders get one extra method: `vBuild()` — short for "validate and build". It acts just like
+`build()` but also throws a `ValidationException` if the message is not valid:
+
+```java
+MyMessage.newBuilder()
+         .setFoo(invalidValue())
+         .vBuild(); // ← Throws ValidationException. 
+``` 
+
+If validation is not required, you may call `build()` or `buildPartial()` to symbolize that 
+the message is likely to be invalid.
+
+The message class also gets an extra method — `validate()`. This method does not throw exceptions.
+Instead, it returns a list of `ConstraintViolation`s:
+
+```java
+MyMessage msg = MyMessage.newBuilder()
+                         .setFoo(invalidValue())
+                         .buildPartial();
+List<ConstraintViolation> violations = msg.validate(); 
+```
+
+If the message is valid, the list is empty. If one or more constraints are violated, all
+the violations will be present in the list.
+
+## Dart validation API
+
+In Dart, we generate functions for validating messages separately from the message classes. Those
+functions can be accessed via the known types:
+
+```dart
+var msg = getMessage();
+var validate = theKnownTypes.validatorFor(msg);
+ValidationError error = validate(msg);
+```  
+
+Similarly to `validate()` method in Java, the validation function does not throw exceptions. A list
+of `ConstraintViolation`s can be obtained from the `ValidationError`.
+
 ## Validation options
 ### Required fields
 
--- a/docs/guides/validation-user-guide.md
+++ b/docs/guides/validation-user-guide.md
@@ -65,9 +65,7 @@ message UserIdentity {
         option (is_required) = true;
         
         EmailAddress email = 1;
-
         GoogleId google = 2; 
-        
         TwitterId twitter = 3; 
     }
 }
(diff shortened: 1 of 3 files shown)


dmdashenkov:
@alexander-yevsyukov, @armiol, PTAL.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,437 @@ headline: Validation User Guide
...
+        GoogleId google = 2; 
+        TwitterId twitter = 3; 
+    }
+}
+```
+
+In this case one of the fields `UserIdentity.email`, `UserIdentity.google`,
+and `UserIdentity.twitter` must be set.
+
+`(is_required)` option is not yet supported in Dart.
```
armiol, line 80: I would have something like a yellow exclamation mark (!) sign here. This isn't particularly related to the story flow.

Or we may add a table with the supported platforms per option at the end of the guide.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,437 @@ headline: Validation User Guide
...
+
+Other cases call for conditional required fields. In particular, some fields of a message must be
```
armiol, line 82: `conditionally`? 

I would rephrase that anyway. `In some other cases, a field may be either required or not, depending on the value of another field. E.g.: ...`

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,437 @@ headline: Validation User Guide
...
+In case if a required field is missing, the validation error message will explicitly say so.
+However, if you need a specific error message for this field, you can provide it via
+the `(if_missing)` option:
+
+```proto
+import "spine/options.proto";
+
+message PhoneNumber {
+    string digits = 1 [(required) = true,
+                       (msg_format = "Phone number must contain digits.")];
```
armiol, line 133: I cannot see any examples of this syntax in our `options.proto`. Is this a valid case?

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,437 @@ headline: Validation User Guide
...
+referenced via any other options.
+
+If `(goes)` option is used, the error message can be customized with the `(goes).msg_format`
+parameter. Note that the message should contain two "`%s`" insertion points: first for the name of
+the field declaring the option and second for the name of the field targeted by the option.
+
+#### When `(required)` is implicit
+
+In Spine, by convention, if the first declared field of a [Command](../introduction/naming-conventions.html#command-definitions),
+an [Event](../introduction/naming-conventions.html#event-definitions), or an entity state is
```
armiol, line 147: "an"? 

Let's rephrase this anyway. `When defining the domain commands, events or entity states, we have found to be convenient that the first field of the respective Message is the identifier. It's just easier to read. Therefore, by convention, Spine treats the first fields of such objects as their IDs: (example).` Then you say — `however if for some reason you have declared the repeated field first, ...`.

Notice, the positive straightforward, then its example, then exceptions and complications.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,437 @@ headline: Validation User Guide
...
+
+```proto
+import "spine/options.proto";
+
+message User {
+    
+    PersonName name = 2 [(validate) = true];
+    // ...
+}
+```
```
armiol, line 199: We will need more usage examples, for now in Java. We could later have tab splits by platform, like Google does in their docs, — when the time comes.

They are much more illustrative along with the explanation, than the explanation alone.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,437 @@ headline: Validation User Guide
...
+    
+    PersonName name = 2 [(validate) = true,
+                         (if_invalid).msg_format = "User name is invalid."];
+    // ...
+}
+```
+
+### Number bounds
+
+For number fields, Spine defines a few options to limit the range of expected values.
```
armiol, line 226: `numeric fields`?

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,461 @@ headline: Validation User Guide
...
 ---
-<p class="coming-soon">Coming soon...</p>
+
+Protobuf is a great tool for structuring domain models. Protobuf definitions help us to standardize
+creating [Value Objects](https://martinfowler.com/bliki/ValueObject.html).
+Spine enhances Protobuf with a validation library.
+
+This guide will walk you though the API of Spine validation library. All of the validation features
+described here are currently supported in the Java environment. Many are supported in Dart as well.
+For more info, see the description of individual constraints.
```
alexander-yevsyukov, line 14: What is supported in JavaScript?

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,461 @@ headline: Validation User Guide
 bodyclass: docs
 layout: docs
 ---
-<p class="coming-soon">Coming soon...</p>
+
+Protobuf is a great tool for structuring domain models. Protobuf definitions help us to standardize
+creating [Value Objects](https://martinfowler.com/bliki/ValueObject.html).
```
alexander-yevsyukov, line 9: Not only value objects. We also define entity states. We also define signals that are immutable, but not value objects per se.

alexander-yevsyukov: I would start with repeating the fact that in Spine projects data model is defined in Protobuf, and which parts of the model are defined using this language. Then say that we need to make sure that a model is consistent with business rules, etc.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,461 @@ headline: Validation User Guide
...
+convention, Spine treats the first fields of such objects as their IDs:
+
+```proto
+import "spine/options.proto";
+
+// The state of the User Aggregate.
+message User {
+    option (entity).kind = AGGREGATE;
+
+    UserId id = 2;
```
alexander-yevsyukov, line 165: The problem with this sample is that we need more imports, and package for `UserId` type. Otherwise, it's going to be misleading for the reader.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,461 @@ headline: Validation User Guide
...
+referenced via any other options.
+
+If `(goes)` option is used, the error message can be customized with the `(goes).msg_format`
+parameter. Note that the message should contain two "`%s`" insertion points: first for the name of
+the field declaring the option and second for the name of the field targeted by the option.
+
+#### When `(required)` is implicit
+
+When defining the domain [Commands](../introduction/naming-conventions.html#command-definitions),
+[Events](../introduction/naming-conventions.html#event-definitions) or entity states, we have found
```
alexander-yevsyukov, line 154: Don't we need a comma before “or entity states”?

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,461 @@ headline: Validation User Guide
...
+field can span between 0 and 59, and the `LocalTime.seconds` field can span between 0.0 and 60.0,
+but can never reach 60. Exclusive boundaries are especially powerful for fractional numbers, since,
+mathematically, there is no hard upper limit which a field value can reach.
+
+Usage of the double dot separator ("`..`") between the bounds is necessary. 
+
+<p class="note">
+In some languages, Protobuf unsigned integers are represented by signed language primitives.
+For example, in Java, a `uint64` is represented with a `long`. If a value of a field in Java will
+overflow into `long` negatives, it will be considered a negative by the validation library. Keep
```
alexander-yevsyukov, line 294: Please capitalise “Validation”, which is the name of the library.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,461 @@ headline: Validation User Guide
...
+a `map` field to only contain non-duplicating elements (values in case of `map`s).
+
+Example:
+
+```proto
+import "spine/options.proto";
+
+// The state of the User Aggregate.
+message User {
+
```
alexander-yevsyukov, line 379: This sample is not a complete definition of the `User` type. So, I think we should have `// ...` in the middle.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,461 @@ headline: Validation User Guide
...
+The definition of `User` itself need not change.
+
+Note that the fields of an external constraint declaration should replicate the fields of the target
+type. In our example, the `Url` type. If the `Url` type had many fields, only those which need any
+validation should be declared. However, note that if the `Url` type declares any validation on its
+own, all of it is discarded and only the "substitute" rules from the `UserPictureConstraint` are
+used.
+
+<p class="warning">
+External constraints are yet supported in Dart.
```
alexander-yevsyukov, line 456: Are NOT yet?

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,461 @@ headline: Validation User Guide
 bodyclass: docs
 layout: docs
 ---
-<p class="coming-soon">Coming soon...</p>
+
+Protobuf is a great tool for structuring domain models. Protobuf definitions help us to standardize
+creating [Value Objects](https://martinfowler.com/bliki/ValueObject.html).
+Spine enhances Protobuf with a validation library.
```
alexander-yevsyukov, line 10: I don't think we extend the language or the library. I'd say that the framework provide the Validation library which builds on top of Protobuf options and plugins to Protobuf compiler.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,461 @@ headline: Validation User Guide
...
+
+This guide will walk you though the API of Spine validation library. All of the validation features
+described here are currently supported in the Java environment. Many are supported in Dart as well.
+For more info, see the description of individual constraints. 
+
```
alexander-yevsyukov, line 15: We need a short overview of the goals of the Validation library. Why do people would want to use it?

Then we need to describe how it's done in general. Then go to options.

docs/guides/validation-user-guide.md
dmdashenkov: We do not have validation in JavaScript.

dmdashenkov:
@armiol, @alexander-yevsyukov, PTAL again.

docs/guides/validation-user-guide.md
```diff
@@ -151,19 +166,21 @@ the field declaring the option and second for the name of the field targeted by
...
 When defining the domain [Commands](../introduction/naming-conventions.html#command-definitions),
-[Events](../introduction/naming-conventions.html#event-definitions) or entity states, we have found
+[Events](../introduction/naming-conventions.html#event-definitions), or entity states, we have found
 to be convenient that the first field of the respective Message is the identifier. Therefore, by
 convention, Spine treats the first fields of such objects as their IDs:
 
 ```proto
 import "spine/options.proto";
+import "spine/core/user_id.proto";
+import "spine/people/person_name,.proto";
```
alexander-yevsyukov, line 176: Please remove the comma before `.proto`.

docs/guides/validation-user-guide.md
```diff
@@ -197,18 +216,31 @@ In order to enable message field checks, use `(validate)` option:
 
 ```proto
 import "spine/options.proto";
+import "spine/people/person_name,.proto";
```
alexander-yevsyukov, line 219: Please remove the comma before `.proto`.

docs/guides/validation-user-guide.md
```diff
@@ -222,13 +254,14 @@ used:
...
```
alexander-yevsyukov, line 257: Please remove the comma before `.proto`.

dmdashenkov:
@alexander-yevsyukov, PTAL again.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,497 @@ headline: Validation User Guide
 bodyclass: docs
 layout: docs
 ---
-<p class="coming-soon">Coming soon...</p>
+
+Spine uses Protobuf for defining domain models. It also helps us to standardize creating
+[Value Objects](https://martinfowler.com/bliki/ValueObject.html). Both those usages require
+something more than just a data structure. One of the commodities required for describing domain
+specifics is data validation. Spine builds on top of Protobuf with a Validation library.
```
armiol, line 11: I would rephrase `Spine builds on top of Protobuf with a Validation library.` with more specifics.

```
Spine extensively uses the Validation library when working with any Protobuf messages.
```

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,497 @@ headline: Validation User Guide
...
+
+The purpose of the Validation library is:
+ - giving framework users an easy way to maintain simple domain invariants;
```
armiol, line 14: Let's extract the common parts of the points.

I also think that mentioning Anticorruption layers and domain invariants is a bit too much. Both these statements require the reader to make several leaps from this guide through their experience towards realizing what the author meant in the first place.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,497 @@ headline: Validation User Guide
...
+described here are currently supported in the Java environment. Many are supported in Dart as well.
+For more info, see the description of individual constraints.
+
+## Validation options
+### Required fields
+
+In Protobuf 2 the concept of required fields used to be built into the language. This proved to be
+a [design mistake](https://stackoverflow.com/a/31814967/3183076) and in Protobuf 3 all the fields
+are optional. However, we've revived the concept in Spine Validation library. Unlike the Protobuf 2
+`required` fields, our `(required)` fields do not effect message serialization and thus can be
```
armiol, line 27: I think `affect`.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,497 @@ headline: Validation User Guide
...
+ignored. Rather than guarding a technical invariant, `(required)` guards a domain model invariant.
```
armiol, line 28: Same as above. There are three unrelated points waiting for the reader to understand themselves:

* Protobuf 2 had this feature, but it was their mistake.
* We do it right. Your serialization/deserialization is safe.
* There are some invariants that need to be guarded. Whether you do that or not — it's your responsibility.

I think you would say this in a much simpler way if you were giving a speech to, let's say, Junior devs.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,497 @@ headline: Validation User Guide
...
+The choice whether or not to use validation lies solely on the developer.
+
+#### How required fields work
+
+Fields in Protobuf may have either a primitive type or a user-defined type. A user-defined type is
```
armiol, line 33: Previously I commented on the flow "simple definition" -> "simple positive example" -> "more details" -> "another example" -> "negative scenarios and don'ts with more technical stuff".

It is still valid as of now.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,497 @@ headline: Validation User Guide
...
+}
+```  
+
+Here, the field `PhoneNumber.digits` is required. If the API user tries to validate an instance of
+`PhoneNumber` without this field, a `ConstraintViolation` is produced:
+
+```java
+PhoneNumber.newBuilder()
+           .setDigits("")
+           .vBuild(); // ← Throws ValidationException.
```
armiol, line 68: I think we should explicitly tell about `vBuild()` and its difference from `build()`.

We describe the API, and this method is surely its part.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,497 @@ headline: Validation User Guide
...
+    uint32 millimeters = 2 [(max) = { value: "1000" exclusive: true }];
+}
+```
+
+#### Ranges
+
+The `(range)` option is a shortcut for a combination of `(min)` and `(max)`. A range specifies both
+boundaries for a numeric field. `(range)` is a `string` option. The `(range)` notation allow
+declaring inclusive and exclusive boundaries. A round bracket ("`(`" or "`)`") denotes an exclusive
+boundary and a square bracket ("`[`" or "`]`") — an inclusive boundary.
```
armiol, line 297: `tells that the boundary is an inclusive one`.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,497 @@ headline: Validation User Guide
...
+    float seconds 3 [(range) = "[0 .. 60.0)"];
+}
+```
+
+In the example above, the `LocalTime.hours` field can span between 0 and 23, the `LocalTime.minutes`
+field can span between 0 and 59, and the `LocalTime.seconds` field can span between 0.0 and 60.0,
+but can never reach 60. Exclusive boundaries are especially powerful for fractional numbers, since,
+mathematically, there is no hard upper limit which a field value can reach.
+
+Usage of the double dot separator ("`..`") between the bounds is necessary.
```
armiol, line 321: I would rather say "is mandatory" or "is required".

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,497 @@ headline: Validation User Guide
...
+    
+    // ...
+
+    spine.time.ZonedDateTime when_placed = 12 [(when).in = PAST];
+    spine.time.ZonedDateTime when_expires = 13 [(when).in = FUTURE];
+}
+```
+
+Note that the value is checked in relation to the current server time. In most cases, this should
+not be an issue. However, be aware that using `FUTURE` in Events and entity states may cause
```
armiol, line 393: This is an absolutely required thing to tell. However, I don't understand why it's bad for immutable Events. And I would tell more about the entity states, as, again, there are some missing pieces for readers between the statement and the consequences.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,497 @@ headline: Validation User Guide
...
+
+```proto
+import "spine/options.proto";
+import "spine/net/email_address.proto";
+
+// The state of the User Aggregate.
+message User {
+    // ...
+
+    repeated spine.net.EmailAddress recovery_emails = 42 [(distict) = true];
```
armiol, line 414: `distinct`

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,497 @@ headline: Validation User Guide
...
+}
+```
+
+### Non-mutable state fields
+
+For entity states, Spine defines a special validation constraint. It is not typically checked in any
```
armiol, line 420: This one does not tell whether I as a framework user am able to use `set_once` for my own routines or not.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,497 @@ headline: Validation User Guide
...
+
+// The state of the User Aggregate.
+message User {
+    // ...
+
+    spine.net.Url profile_picture = 42;
+}
+```
+
+How do we add validation to the fields inside `Url` so that only the `User.profile_picture` is
```
armiol, line 462: Let's also show how `Url` looks. We don't expect readers to take their time and go look into the definition themselves, do we?

It makes even more difference, given you speak of `Url` structure further, implying it's a well-known thing.

dmdashenkov:
@armiol, PTAL again.

docs/guides/validation-user-guide.md
armiol: Can I as a Spine user utilize the `(set_once)` myself? If so, then start with what it is and how to use it. And then describe how Spine uses it internally. This is a piece of advanced information, not essential to gain understanding.

docs/guides/validation-user-guide.md
```diff
@@ -4,4 +4,615 @@ headline: Validation User Guide
...
+In Java, you can validate messages against a `set_once` constraint manually. For that, pass the old
+version and the new version of a message to the `Validate.checkValidChange()` method. For example:
+
+```java
+MyMessage old = getMessage();
+MyMessage changed = doSomeStuff(old);
+Validate.checkValidChange(old, changed);
+```
+
+In Dart, there is no support for this feature.
```
armiol, line 543: Let's mark this paragraph the same way we did earlier.

docs/guides/validation-user-guide.md
dmdashenkov: We only capitalize "Validation" when in combination with "library" as in _"Validation library"_. The process "validation" is not capitalized.

dmdashenkov:
@armiol, PTAL again.

I think two first paragraphs describe the motivation behind creating the library as well as the goals pursued by the user. Is there anything else I should add?

Issue #171, opened 2019-04-15
Update Naming Conventions Guide


meizabel:
- Current guide contents should be revised to reflect our current approach to DDD and recent Spine framework enhancements (a suggestion by @armiol ). 
- Based on the Naming Conventions guide review comment by @alexander-yevsyukov: 
One more issue we need to address within the framework of this one Command definitions in `commands.proto` file: we need to mention a name of a file. Otherwise there is a small chance of confusing the name of a file with its content: “Do I put this text at the end of the file?”

PR #270, opened 2020-04-13
Improvements for naming conventions

alexander-yevsyukov:
This PR:
  * Updates the table of the contents of the Naming Conventions document (addressing #171).
  * Adds recommendation for excluding the generated code from VCS.

Other improvements regarding improving the structure of the documentation will be addressed in the following PRs.


docs/introduction/naming-conventions.md
```diff
@@ -25,14 +25,36 @@ and others. You can find similar cases in the framework API which has `EventId`,
...
 We find it convenient to define ID types in one file called `identifiers.proto`. 
-A typical project is likely to have more than one Bounded Context, thus you will have several
+A typical project is likely to have more than one Bounded Context. Thus, you will have several
 `identifiers.proto` files. 
 Each of them resides under the directory with proto files defining the data model of the
-corresponding Bounded Context.   
+corresponding Bounded Context. For example:
 
-### Command definitions
+```
```
alexander-yevsyukov, line 33: @JuliaEvseeva, I need to highlight `identifiers.proto` in this fragment. I cannot do it using the markdown syntax, which is used now.  I can do it via `<pre>`, `<code>, and `<strong>`, but it does not turn on color highlighting, which is available for other code fragments. When I put classes that are used in other code fragments (including nesting of corresponding `<div>`s) resulting page starts displaying those tags. Please suggest how I can highlight parts within a code block.

JuliaEvseeva: Now we can do it like this:
```
<div class="highlighter-rouge">
<pre class="highlight">
<code>
    myproject/
      users/
        src/
          main/
            java/
            proto/
              user.proto
              group.proto
              ...
              <span class="kd">identifiers.proto</span>
      tasks/
        src/
          main/
            java/    
            proto/
              task.proto
              project.proto
              ...
              identifiers.proto
        ...
</code>
</pre>
</div>
```

alexander-yevsyukov: Thanks. It worked. Strangely, it didn't work for me when I tried the same. Presumably, because of the wrong nesting of tags.

JuliaEvseeva: Unfortunately, this method will not work if we want to define a language for the code. I'll think about how to make it more universal.

PR #272, opened 2020-04-15
Improve the About page

alexander-yevsyukov:
This PR updates the About page in the following ways:
  * Features section was removed. We have features on the home page. Let's DRO.
  * Added the Principles section. It is the first step for removing the Motivation page from the Introduction.
  * Consolidated tools and online systems into one section.
  * Moved libraries into a separate section updating the text. We need to have a separate section for Clients, so Libraries got light blue background color.
  * Clients come in a separate section with updated title.
  * Added the call to action to the bottom.

### DIFF
--- a/about/index.html
+++ b/about/index.html
@@ -31,17 +31,17 @@ headline: 'About Spine'
     </div>
 </section>
 
-<section class="story-section">
+<section class="principles-section">
     <div class="container-fluid content-holder">
         <h2 class="bold section-title">Principles</h2>
         <div class="row">
             <div class="col-sm-7">
                 <p>These are the principles our framework is based on:</p>
-                <div class="principles">
+                <div>
                     <ul>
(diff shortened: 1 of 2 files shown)


alexander-yevsyukov:
@JuliaEvseeva, I cannot make the Principles text wrap in the same way (on smaller screens) as the intro section does. Can you help with this, please?

about/index.html
```diff
@@ -31,42 +31,37 @@
...
-        {% include feature-section.html %}
-    </div>
-</section>
-
-<section class="story-section">
-    <div class="container-fluid content-holder">
-        <h1 class="bold section-title">The story behind Spine</h1>
+        <h2 class="bold section-title">Principles</h2>
         <div class="row">
             <div class="col-sm-7">
```
JuliaEvseeva, line 38: To make the same Principles text wrap as the intro section does, we need to change the class `col-sm-7` to `col-lg-7`.

PR #280, opened 2020-05-21
Implement Algolia search

JuliaEvseeva:
This PR brings a new Algolia search implementation.

The navigation and search input updates:
- the navigation styles were cleaned up to avoid Algolia styles overriding;
- the navigation search was moved to a new component `_includes/nav-search.html` and corresponding styles were also moved to the `_nav-search.scss`;
- the search input was moved from the hamburger menu to the navigation menu on mobile devices (it is needed for the correct Algolia dropdown work that cannot be opened inside the mobile navigation holder);
- now the search input will be opened in the full width of the menu on mobile devices;
- the input work on the iOS devices was fixed.


JuliaEvseeva:
@andreySadomovskiy, PTAL.

PR #291, opened 2020-06-05
Fix broken navigation and footer links

JuliaEvseeva:
This PR fixes broken links in mobile navigation and footer.

Fixed links:
- `Quick Start`;
- `Guides`;
- `Tutorials` link has been removed.

Also, the `Overview` page will be active on UI, if it is the current page.

### DIFF
--- a/_data/navigation/doc_side_nav.yml
+++ b/_data/navigation/doc_side_nav.yml
@@ -1,6 +1,6 @@
 doc_side_nav:
   - page: Overview
-    url: /docs/index.html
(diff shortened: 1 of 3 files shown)


JuliaEvseeva:
@alexander-yevsyukov, in mobile navigation and footer, we have the `Tutorials` and` Reference` links. Now they are broken. Maybe we can hide them for now or change to something.

JuliaEvseeva:
@alexander-yevsyukov, PTAL.

PR #293, opened 2020-06-05
Add `API Reference` page as `index` page

JuliaEvseeva:
In this PR I have moved all the `API Reference` links from the side navigation to the separate page `/docs/reference/index.md`.


JuliaEvseeva:
@alexander-yevsyukov, PTAL.

PR #299, opened 2020-06-08
Advances Quick Start guide further

alexander-yevsyukov:
This PR:
  * Puts the Quick Start guide into a sub-directory and updates the navigation accordingly.
  * Improves wording for commands and text flow between sections and to code samples.
  * Adds missing text for `java_multiple_files` option.
  * Adds intro text for events.


alexander-yevsyukov:
@JuliaEvseeva, please see that I've updated navigation correctly.

docs/quick-start/index.md
```diff
@@ -168,7 +168,7 @@ We are not reviewing these parts of the project configuration deeper because the
 related to the use of the Spine framework. If you're interested in more details, please look into
 the code of these scripts.
 
-Now, let's look into the definition of the data structures of the Hello context. 
+Now, let's look into the data structures of the Hello context.
```
armiol, line 171: Maybe `structure` would read better? There is also `The data of the Hello context are defined ...` down below. 

Both `structures` and `the data are defined` seem to be examples of a way-too-literal translation from Russian.

docs/quick-start/index.md
```diff
@@ -264,10 +273,57 @@ The second field is marked as `(required)` using the custom option imported in
 the `spine/options.proto` file above. This command does not make much sense if there is no text
 to print.
 
+<p class="note">In Protobuf a data type is either a **`message`** (we can send it) or an **`enum`**.
```
armiol, line 276: I would say `In a majority of cases, ...`. Because there is also options themselves. Eventually, the readers might want to have a look a `options.proto` which they imported for so many times. Each structure starts with `extends ...Options` in it.

_includes/footer.html
```diff
@@ -12,7 +12,7 @@
           </div>
           <ul class="footer-nav-list collapse" id="footerDocList">
             <li><p><a href="{{ site.baseurl }}/docs/introduction">Introduction</a></p></li>
-            <li><p><a href="{{ site.baseurl }}/docs/quickstart-java.html">Quick Start</a></p></li>
+            <li><p><a href="{{ site.baseurl }}/docs/quickstart/">Quick Start</a></p></li>
```
JuliaEvseeva, line 15: Let's have `quick-start` instead of the `quickstart` to be according to the folder name.

docs/quick-start/index.md
alexander-yevsyukov: I want to keep it simple for the novice readers. That's why I make it straight as much as possible. Also, options are meta-data. So, an option type is meta-data type.

PR #300, opened 2020-06-09
More quick start

alexander-yevsyukov:
This PR adds text for server-side data of the Quick Start guide.

### DIFF
--- a/docs/quick-start/index.md
+++ b/docs/quick-start/index.md
@@ -187,7 +187,7 @@ the following files:
     suffix in their names.</p>
 
 These two files define signals used by the Hello context. There's also data of the `Console`
-Process Manager, which is defined in the package `server` in the file `console.proto`.
+Process Manager, which is defined in the package **`server`** in the file **`console.proto`**.
 
 <p class="note">
 We arrange the sub-package `server` to highlight the fact that this is server-only data. It is not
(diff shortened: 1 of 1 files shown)


docs/quick-start/index.md
```diff
@@ -316,13 +316,68 @@ The sole event in this project is declared this way:
...
+    // The login name of the computer user.
+    string username = 1;
+
+    // Text lines of the screen.
+    repeated string lines = 2;
+}
+```
+
+The option `(entity)` tells us that this type is going to be used by a `ProcessManager`.
+The first field of the type is not marked as `(required)` because the framework assumes it holds
```
armiol, line 359: ```The first field of the type holds an ID of this entity. The framework assumes such fields as implicitly `(required)`.```

docs/quick-start/index.md
```diff
@@ -316,13 +316,68 @@ The sole event in this project is declared this way:
...
+an ID of the entity. Then goes the declaration of the remote screen output. This field is not
```
armiol, line 360: ```The value of this field is empty until something is printed on the screen. Therefore, it is not marked `(required)`.```

docs/quick-start/index.md
```diff
@@ -316,13 +316,68 @@ The sole event in this project is declared this way:
...
 }
 ```
 
+The event tells which text was printed for a user. Both of the fields are marked as `(required)`
+because the event does not make much sense if one of them is empty. 
+
+<p class="note">
+Unlike for commands, the framework does not assume that the first event field is <em>always</em>
+populated. This is so because default routing rules for commands and events are different. 
+Routing for events start with using an ID of the entity which emitted the event.
```
armiol, line 332: `starts`?

I would split this thought a bit differently:
```
When an event is produced by some entity, it remembers the ID of this producer entity. 
By default, the framework uses the producer ID to route events to their target entities — if they have identifiers of the same type. 
If the type of producer ID does not match one of the target entity, then event fields are analyzed.
It is also possible to set custom routing rules.

PR #312, opened 2020-06-16
Update copyright year

JuliaEvseeva:
This PR brings the updated copyright year.
2015 — is the year when repositories became visible.


JuliaEvseeva:
@andreySadomovskiy, PTAL.

PR #315, opened 2020-06-22
Prepare screen size variables to easy switch to mobile search

JuliaEvseeva:
This PR prepares variables to make it easy to define a screen size when the search input should be opened on mobile.

This is useful when we start adding new pages to the main navigation.


JuliaEvseeva:
@andreySadomovskiy, PTAL.

PR #318, opened 2020-06-26
Update Identifiers naming convention

yuri-sergiichuk:
This PR brings in the new way of naming IDs within the proto messages.


yuri-sergiichuk:
@alexander-yevsyukov PTAL

PR #330, opened 2020-07-14
Add subscription info to the Release Notes

JuliaEvseeva:
This PR brings subscription information to the Release Notes.

This information block contains a text to `Subscribe to @SpineEngine on Twitter`. And it will be visible at all “release notes” pages.

Also, common variables were cleaned up and added to the `config.scss` file.


JuliaEvseeva:
@andreySadomovskiy, PTAL.

PR #336, opened 2020-07-16
Remove `release-notes-feed.xml` file

JuliaEvseeva:
This PR removes the `release-notes-feed.xml` file.
The file was used to create an RSS feed to make a subscription.

The use of collections for the Release Notes no longer makes sense. Therefore, it was redone back into the simple folder.
Thus, we get rid of duplicate folders and keep everything in one.

Also, favicon config files were moved to the root folder to make them work on the old IE versions.


JuliaEvseeva:
@andreySadomovskiy, PTAL.

PR #359, opened 2020-08-07
Update Jekyll and other dependencies

alexander-yevsyukov:
This PR fixes the Jekyll problem with leakage of the `embed-code` instructions into the content of the pages. 
It was discovered when visiting https://spine.io. The problem manifested in both Quick Start and Integration guides.

The Jekyll version was updated to 3.9.0 to reproduce the bug locally. It did not manifest itself with the previous version.

Since GitHub renders the site itself using Jekyll (we do not submit the `_site` directory to version control), most likely they updated Jekyll recently. That's why we observe the problem when the site is published.

`embed-code` instructions were commented out, not removed. They need to be restored after the Jekyll / GitHub Pages issue is fixed.


alexander-yevsyukov:
@oleksandra-kovalenko, please find the issue with Jekyll or GitHub Pages which caused the problem which this PR addressed. 

If there is no such issue, please create it and link it to this PR.

alexander-yevsyukov:
[This Jekyll issue](https://github.com/jekyll/jekyll/issues/8315) relates to support of Kramdown v2 syntax. The description of the issue mentions “invisible errors” that may occur to GitHub Pages users because if Pages migrate to Jekyll v4.

I suspect, we're dealing with one of such “invisible” issues...

alexander-yevsyukov:
Here are the [details](https://github.com/jekyll/jekyll/pull/8323/files#diff-667116299da3eaf26887e45eb83e39ad) on the Jekyll 3.9.0 release.

alexander-yevsyukov:
Here's the link to [Kramdown news](https://kramdown.gettalong.org/news.html). The post for v2 says it has breaking changes.

@armiol, FYI.

alexander-yevsyukov:
The version 2.3.0 of Kramdown introduces the following breaking change (quote from the news):

```
Although this is a minor version bump there is one breaking change: Parsing of XML processing instructions was removed because they are invalid for HTML5 documents.
```

Here's the link to the [issue](https://github.com/gettalong/kramdown/issues/660).

alexander-yevsyukov:
The [Kramdown issue](https://github.com/gettalong/kramdown/issues/660) which affects us is already closed.

PR #370, opened 2020-08-18
Gradle user guide

dmdashenkov:
In this PR we add a user guide for using Spine Gradle plugins: Bootstrap, Model Compiler, and ProtoJs.

In the guide, the user is introduced to the basics of configuring a Spine-based project beyond a "Hello World" app. Such usecases may include working with multiple co-dependant Bounded Contexts, maintaining a client-server CQRS communication, etc.

In this PR we also replace hardcoded references to API docs and Git repos with Jekyll variables.


dmdashenkov:
@armiol, PTAL.

docs/guides/gradle.md
```diff
@@ -0,0 +1,166 @@
...
+
+You can also find this declaration on the [Gradle Plugin Portal](https://plugins.gradle.org/plugin/io.spine.tools.gradle.bootstrap),
+or on our [Getting Started page]({{site.baseurl}}/docs/quick-start).
+
+## Spine Bootstrap plugin 
+
+Spine Bootstrap plugin (Bootstrap for short) can configure a Spine Java server and JS client for
+you.
+
+The idea is to use Gradle subprojects for different parts of the system.
```
armiol, line 34: Let's connect this sentence to the previous one and the content which follows after. 

Something like this:

```
We recommend having separate Gradle subprojects for domain model definition, server 
implementation, and client code. Bootstrap is applied a bit differently in each
of these cases.
```

Also, the previous sentence says nothing about the model. Maybe we could rewrite it to give a more generic definition for the plugin per se:

```
Spine Bootstrap plugin (Bootstrap for short) serves to automate the configuration of the modules
in your Spine-based app.
```

docs/guides/gradle.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Model definitions
+
+One subproject would contain Protobuf definitions for the domain model. This subproject is
```
armiol, line 38: I would say

`Let's assume one of the subprojects contains...`

docs/guides/gradle.md
```diff
@@ -0,0 +1,166 @@
...
+typically called `model`. If the system contains more than one Bounded Context, there could be
```
armiol, line 39: Let's first say how to do that for a single subproject, in detail.

Then, let's mention there may be several contexts. And how to wire them up etc.

docs/guides/gradle.md
```diff
@@ -0,0 +1,166 @@
...
+```
+If one of the Bounded Contexts shares some domain language with another, add a dependency between
+them. This way, the downstream context may use Protobuf definitions of the upstream context.
+```groovy
+dependencies {
+    implemetation(project(':model-users'))
+}
+```
+
+### Java implementation
```
armiol, line 53: `Java server implementation`?

docs/guides/gradle.md
```diff
@@ -0,0 +1,166 @@
...
+
+Each Bounded Context should have a separate Gradle subproject for implementing server-side
```
armiol, line 55: The idea of speaking of multiple contexts is that a reader obtains a more complex picture in her mind. And the problem here is that it's not likely what a first-learning reader starts from. I mean, even if they have multiple contexts, they are not going to deal with many of them right away knowing nothing of the Spine and the Gradle configuration.

When learning how to do things, it's always a good idea to show a simple and streamlined flow. And then have a more complex abstraction over it. Applying to this article, I would describe how things work for a single bounded context first. And then I would have a separate section (or maybe even another article), where the problem of inter-dependencies would be covered — as it ain't a simple thing at all, it's not just the dependency configuration.

docs/guides/gradle.md
```diff
@@ -0,0 +1,166 @@
...
+```groovy
+spine.enableJava().server()
+
+dependencies {
+    protobuf(project(':model'))
+}
+```
+This will add dependencies to the `spine-server` artifact and set up code generation for
+the domain model types. Note that it is perfectly normal to have more Protobuf types in
+these modules, as long as those types are internal to the server and are not a part of
```
armiol, line 67: This is a tricky part. As a first-time reader, would you want to learn about the existence of "internal" Proto types from this article? I think, not.

docs/guides/gradle.md
```diff
@@ -0,0 +1,166 @@
...
+
+Those plugins are Spine Model Compiler for Java subprojects and Spine ProtoJs plugin for JavaScript
+submodules. Under the hood, Bootstrap uses those plugins to do the work. This means that Bootstrap
+automatically applies the correct low-level plugin for you.
+
+### Model Compiler
+
+Spine Model Compiler is a Gradle plugin which executes all the code generation routines via several
+Gradle tasks as well as the `modelCompiler { ... }` extension, which allows you to configure those
+tasks The tasks are:
```
armiol, line 140: Judging on the "verbose" word, this section should speak of a fine-tuning for the model compiler. However, to me, it looks more like a quick reference.  As a reader, what I would want to do with this knowledge? Do I ever use these tasks, and if yes, then when and why?

We should either give some examples or just move this content to the `README.md` of the respective module and just provide readers with a link.

docs/guides/gradle.md
```diff
@@ -0,0 +1,166 @@
...
+
+ - `preClean` — deletes directories generated by Spine if `clean` task is also executed. Append
+   directories to `modelCompiler.dirsToClean` to delete some other dirs which could be created
+   during code generation. 
+ - `mergeDescriptorSet` & `mergeTestDescriptorSet` — generate a merged Protobuf type descriptor set
```
armiol, line 145: If we speak "[task] deletes" in the `preClean` section, we should speak "generates" etc in the following sections as well.

armiol (changes requested):
@dmdashenkov please see my comments.

dmdashenkov:
@armiol, PTAL again.

docs/guides/gradle.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Model definitions
+
+Let's assume one of the subprojects contains Protobuf definitions for the domain model. This
+subproject is typically called `model`. In `build.gradle` for that subproject, declare:
+```groovy
+spine.assembleModel()
+```
+This way, the subproject will receive all the required Spine dependencies but no code will be
+generated from Protobuf.
```
armiol, line 45: Let's discuss the original idea of this section tomorrow in person. I don't fully understand what you mean here :(

docs/guides/gradle.md
```diff
@@ -0,0 +1,171 @@
...
+</p>
+
+### Java web server
+
+If your project contains a JavaScript frontend, you may declare a `web-server` subproject, which
+processes the HTTP requests from JS. In `web-server/build.gradle`:
+```groovy
+dependencies {
+    implementation("io.spine:spine-web:${spine.version()}")
+    implenemtation(project(':server'))
```
armiol, line 95: Please fix a typo.

Ideally, we would want to refer to some code in a real project instead. Not sure it exists now, but can we use `todo-list`?

armiol (commented):
@dmdashenkov this is much-much better than the last time. I'd like to discuss one thing (see my comment) with you though.

dmdashenkov:
@armiol, PTAL again. If needed let's discuss if the provided explanations are detailed enough.

docs/guides/gradle.md
```diff
@@ -0,0 +1,177 @@
...
+If your server should be deployed as a whole, use a single `web-server` for all the contexts. If you
+would like to deploy different contexts separately, declare a specific `web-server` subprojects
+for each of those contexts. See the [API reference]({{site.core_api_doc}}/server/io/spine/server/ServerEnvironment.html#use-io.spine.server.transport.TransportFactory-io.spine.base.EnvironmentType-)
+for the instructions on integrating Bounded Contexts. Also, see [this guide]({{site.baseurl}}/docs/guides/integration.html)
+on the principles of integrating separate Bounded Contexts and third-party systems in Spine.
+
+## Verbose configuration
+
+If the Bootstrap configuration is not customizable enough for you, there are other Gradle plugins
+which may provide the right API.
```
armiol, line 154: Maybe `the fine-grained API`?

armiol (approved):

PR #372, opened 2020-08-19
Home page code

alexander-yevsyukov:
This PR updates the code samples at the Home page. In particular:
  * All the samples now revolve around one command, one event, and one aggregate state. The state of a projection which used to be next to the aggregate state is dropped to make code samples for focused.
  * A code sample of the command handing test is added after the business logic code sample to follow the text of the page.
  It also gives the visitor instant idea on what it's like to work with the projects based on the framework.
  * Added code sample of the `ServerEnvironment` configuration.
  * Added a “sample” of Gradle command line related to deployment. It's just a finishing touch made with a thing related to the code.

This PR adds embedded samples as Proto and Java code under `_code/samples`, but they were embedded into the page manually because of `embed-code` does not handle a sophisticated scenario of embedded we need in this case.


andreySadomovskiy:
@armiol , PTAL

armiol:
@alexander-yevsyukov @andreySadomovskiy LGTM.

armiol (approved):

PR #376, opened 2020-08-27
Show the tooltip only on the disabled button

JuliaEvseeva:
This PR shows the tooltip only on the disabled button and hides it on the enabled button.

### DIFF
--- a/js/pricing.js
+++ b/js/pricing.js
@@ -93,13 +93,14 @@ $(
          * If `true` — enables the element, otherwise — disables
          */
         function changeElementState(element, enable) {
+            const button = element.find($orderButton);
             if (enable) {
-                element.find($orderButton).removeAttr('disabled');
-                element.find($orderButton).removeClass('disabled');
+                button.removeAttr('disabled');
+                button.removeClass('disabled');
                 element.removeAttr('data-original-title');
             } else {
-                element.find($orderButton).attr('disabled', true);
-                element.find($orderButton).addClass('disabled');
+                button.attr('disabled', true);
+                button.addClass('disabled');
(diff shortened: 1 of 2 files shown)


JuliaEvseeva:
@andreySadomovskiy, PTAL.

js/pricing.js
```diff
@@ -85,16 +89,19 @@ $(
          * Changes element's disabled/enabled state.
          *
          * @param {jQuery} element the element to change the state for
+         * @param {jQuery} elementHolder the element holder to apply or hide a tooltip
          * @param {boolean} enable denotes whether the element should be enabled.
          * If `true` — enables the element, otherwise — disables
          */
-        function changeElementState(element, enable) {
+        function changeElementState(element, elementHolder, enable) {
```
andreySadomovskiy, line 96: I'd rather use only `element` parameter, but pass to the function the `orderButtonHolder `. And then find the button inside.

andreySadomovskiy (commented):
@JuliaEvseeva, please  PTAL at the comment below.

js/pricing.js
```diff
@@ -90,11 +94,13 @@ $(
          */
         function changeElementState(element, enable) {
             if (enable) {
-                element.removeAttr('disabled');
-                element.removeClass('disabled');
+                element.find($orderButton).removeAttr('disabled');
```
andreySadomovskiy, line 97: Use `const` for the `element.find($orderButton)`, please.

andreySadomovskiy (approved):
@JuliaEvseeva , PTAL at the comment bellow and LGTM.

PR #377, opened 2020-08-31
Update examples

dmdashenkov:
In this PR we update the example projects, so that they can be built without a hardcoded Spine Protoc plugin JAR.


dmdashenkov:
@alexander-yevsyukov, PTAL.

alexander-yevsyukov (approved):
LGTM

PR #388, opened 2020-09-07
Fix checkboxes when hitting the `back` button

JuliaEvseeva:
This PR fixes the situation when checkboxes are still checked but the button is disabled when hitting the `back` button.
Also, open the privacy pages in the new browser tab to avoid misunderstanding how to back to the page.


JuliaEvseeva:
@andreySadomovskiy, PTAL.

andreySadomovskiy (approved):
@JuliaEvseeva , LGTM

PR #402, opened 2020-10-16
Hide the default Webkit `X` icon from the search input

JuliaEvseeva:
This PR hides the default WebKit `X` icon from the search input.

The default icon doesn't suit us in the style. It is dark and not visible on the blue background. If we try to add a custom icon it has some functionality problems.


JuliaEvseeva:
@andreySadomovskiy PTAL.

andreySadomovskiy (approved):
@JuliaEvseeva , LGTM

PR #405, opened 2020-12-01
Make scripts friendlier to custom setups

yuri-sergiichuk:
This PR improves our helper scripts setup allowing users with different Bash setups to use them without being forced to create custom configuration files.

It is usually a preferred way to rely on the user's `env` configuration over hard-coding the `/bin/bash` shebang. Also, users may not use `bash` as their default shell, then they will not have the bash profile file available. The last change is the usage of the POSIX-compatible [dot command](https://www.shell-tips.com/bash/source-dot-command/) over the Bash-specific `source` command to make the script a bit more friendly to other shells.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -41,6 +41,11 @@ LICENSE text
 
 # Declare files that will always have LF line endings on checkout.
 *.sh text eol=lf
+_script/jekyll-build eol=lf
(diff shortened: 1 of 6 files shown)


alexander-yevsyukov (approved):
LGTM

PR #428, opened 2021-09-05
Fix typo in quick start doc

tripolkaandrey:
I came across a minor typo in [Getting started with Spine in Java](https://spine.io/docs/quick-start/) documentation. Here is the fix.

Before:
<code>
The first field of the type holds <b>and</b> ID of this entity
</code>
After:
<code>
The first field of the type holds <b>the</b> ID of this entity
</code>


oleksandra-kovalenko (approved):
LGTM

PR #442, opened 2022-01-18
Make code blocks wider

### LINKED ISSUE #433: Make text blocks wider
Our code conventions allow for 100 columns in a source code line. We should not go to _that_ length in our code examples, but the “window” for code content we have now at the site is too narrow.

First of all, the central content block of the documentation section already limits us in terms of width. In addition to this, we have margins around the code blocks _and_ margins inside the code blocks. As the result, when we need to show more or less wide source code, we have to figure out custom code wrapping tricks.

There's a relatively simple solution to make the code blocks wider. See how Protobuf documentation [does it](https://developers.google.com/protocol-buffers/docs/reference/javascript-generated). The code blocks occupy the whole width of the content panel. The source code indented as the main text, which makes it easier to read.

In addition, this layout give some space at the top right corner or the block so that switching between programming languages (e.g. Java and Kotlin) can be one “in-place”.

JuliaEvseeva:
This PR closes #433.

This PR brings several changes to the block code displaying:
- The code blocks are now full-width inside the content area in the documentation.
- The code tabs are now in Material style.
- The icon, which switches the code theme color, is now placed near each code block.
- The global color selector has been removed.

https://user-images.githubusercontent.com/22611365/150313797-76dfe449-ecf8-43ff-818b-06d2e450e3a3.mov


JuliaEvseeva:
@alexander-yevsyukov PTAL

alexander-yevsyukov (approved):
LGTM

PR #447, opened 2022-02-10
Remove Google Universal Analytics script

andreySadomovskiy:
The Google Universal Analytics script was removed in this PR, because we will use Google Tag Manager for properties management.


JuliaEvseeva (approved):
@andreySadomovskiy LGTM

PR #452, opened 2022-02-16
Fix duplicated results in DocSearch

JuliaEvseeva:
This PR makes all links consistent throughout the site.

The problem is that many links lead to the same page.
For example: `/docs`, `/docs/`, `/docs/index`, `/docs/index.html`. 
This causes a big problem for DocSearch because it analyzes all the links on the site and creates records for each one.

Screenshot 2022-02-16 at 13 51 31

So, the best way to avoid duplication is to stick to the URLs that are in the [`_data/navigation/doc-side-nav.yml`](https://github.com/SpineEventEngine/SpineEventEngine.github.io/blob/master/_data/navigation/doc_side_nav.yml).


JuliaEvseeva:
@andreySadomovskiy PTAL

andreySadomovskiy (approved):

PR #457, opened 2022-04-26
Resolve Dependabot alerts

andreySadomovskiy:
This PR updates dependencies.


oleksandra-kovalenko:
proof-links warnings are shooting for dddi.dev, which is built on russian-based Tilda, which is currently partially available in some countries.

PR #472, opened 2023-11-23
Improve dependency reports

armiol:
In scope of this changeset, a few things were improved in relation to displaying the third-party dependencies along with their licenses:

1. New `dependencies.md` file is now loaded from each repository instead of `license-report.md` as the latter is being renamed due to [config#498](https://github.com/SpineEventEngine/config/issues/498). For those repositories, in which such a renaming is not done yet, `license-report.md` is still loaded as a fallback.

2. In reports, the repeating sections telling about dual-licensing and crediting the Gradle plugin for dependency reporting are moved from per-artifact blocks to the top of [this page](https://spine.io/oss-licenses/). Also, we don't display the generation time for each artifact, as it may be fetched from its publication date anyway.

3. Spine Users report was removed, as this library is not in active development for a few years already.

Here is a screenshot of locally rendered page with all the changes applied:

Screenshot 2023-11-23 at 15 55 09

### DIFF
--- a/js/oss-licenses.js
+++ b/js/oss-licenses.js
@@ -68,7 +68,7 @@ $(
             });
 
             /**
-             * Makes all markdown links external.
+             * Makes all Markdown links external.
              */
             linkElements.addClass('external');
             linkElements.attr('target', '_blank');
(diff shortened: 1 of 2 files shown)


armiol:
@JuliaEvseeva PTAL.

JuliaEvseeva (approved):

PR #474, opened 2024-03-14
Use an absolute URL for sitemap in `robots.txt`

armiol:
Currently, a sitemap syntax in our `robots.txt` is incorrect, as it refers to the `sitemap.xml` file via a relative URL.

According to [Google's recommendations](https://developers.google.com/search/docs/crawling-indexing/robots/create-robots-txt), the URL must be absolute. This PR does just that.


alexander-yevsyukov (approved):
LGTM

PR #475, opened 2024-03-18
Exclude `/dart/` from crawling

armiol:
Previously, there were several thousand `404` responses detected by Google Search Console review process, the majority of which were related to the generated documentation for Spine Dart client.

Attempts to address the problem, or at least, locate its roots failed. Seems like some previously crawled pages are still in Google Search cache, but now (as of `1.9.0` release) they are missing.

Therefore, within this PR, the whole `/dart/` subtree is excluded from crawling altogether. Hopefully, it will enforce the cache cleanup.

In some future, we may re-enable the crawling for this part of the site. Most likely, it will make sense to do together with 2.0.0 release.

### DIFF
--- a/robots.txt
+++ b/robots.txt
@@ -4,7 +4,6 @@ layout: null
 
 User-agent: *
 {% for node in site.pages %}{% if node.noindex %}{% assign isset = true %}Disallow: {{ node.url }}
-{% endif %}{% endfor %}{% if isset != true %}Disallow:
-{% endif %}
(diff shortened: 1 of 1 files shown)


armiol:
@alexander-yevsyukov PTAL.

armiol:
See #476 for the issue on re-enabling.

alexander-yevsyukov (approved):
LGTM

PR #536, opened 2026-04-23
Add checkout page

Vladyslav-Kuksiuk:
This PR adds checkout page `.../checkout?product={product-id}`.

Predefined test Paygate product: `spine-standard-support`.

### Page view:
image
image

### Server Error view (when something is broken in the middle of checkout):
image

### Product not found page (when Paygate is available, but answers 404):
image

### Thank you page (`.../checkout-completed`)

image

### Ooops page

image


site/assets/js/pages/checkout/index.js
```diff
@@ -0,0 +1,228 @@
+'use strict';
```
MykytaPimonovTD, line 27: copyright

site/assets/js/pages/checkout/index.js
```diff
@@ -0,0 +1,228 @@
...
+
+import * as params from '@params';
+import {createPurchaseClient} from '../../modules/paygate/purchases';
```
MykytaPimonovTD: Use absolute path everywhere.

site/assets/js/pages/checkout/phone-codes.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
```
MykytaPimonovTD, line 27: copyright. Check everywhere

MykytaPimonovTD:
I propose delete circle with '!' from this view. @JuliaEvseeva what do you think?

Screenshot 2026-04-27 at 14 35 51

site/content/checkout-completed/index.md
```diff
@@ -0,0 +1,5 @@
+---
+description: Your checkout has been completed successfully.
+body_class: checkout-page
```
MykytaPimonovTD, line 4: disable from sitemap.

site/content/checkout-completed/index.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
```
MykytaPimonovTD: redundant line.

site/content/checkout-completed/index.md
```diff
@@ -0,0 +1,5 @@
+---
+description: Your checkout has been completed successfully.
```
MykytaPimonovTD: Maybe we need a `title` frontmatter?

site/content/checkout/index.md
```diff
@@ -0,0 +1,7 @@
+---
+description: Checkout
```
MykytaPimonovTD: Looks like title, not description.

site/layouts/_partials/components/result-panel.html
```diff
@@ -0,0 +1,31 @@
+{{ $title := .title }}
```
MykytaPimonovTD, line 29: copyright and documentation.

site/layouts/404.html
```diff
@@ -0,0 +1,27 @@
+{{ define "main" }}
+    {{ partial "components/navbar/navbar.html" . }}
+    {{ partial "components/nav-hero.html" (dict "Params" (dict "hide_page_headline" true) "Title" "") }}
```
MykytaPimonovTD: First of all, it is not a `hide_page_headline` property. It just does nothing.

Secondly, If you do not what to display hero, it is better just do not include hero to view. So, simply delete this line.

site/layouts/_partials/components/result-panel.html
```diff
@@ -0,0 +1,31 @@
...
+{{ $titleTag := .titleTag | default "h2" }}
```
MykytaPimonovTD: It is strange property. Let's always use 'h1' or use bool prop 'title_as_h1'  like in 'nav-hero.html'.

site/layouts/_partials/components/result-panel.html
```diff
@@ -0,0 +1,31 @@
...
+{{ $markText := .markText | default "" }}
+{{ $markClass := .markClass | default "result-panel-mark-success" }}
```
MykytaPimonovTD: Hide mark to object. It is dict with 2 params: `text` (or `content`) and `icon_class`.

If this prop is missing do not render a following line:

```html
<div class="result-panel-mark {{ $markClass }}" aria-hidden="true">{{ $markText }}</div>
```

JuliaEvseeva:
> I propose delete circle with '!' from this view.

Yes, we don’t need this circle here. Also, there is an extra space before the period in the message.

site/layouts/_partials/components/result-panel.html
```diff
@@ -0,0 +1,31 @@
...
+{{ $lines := .lines | default slice }}
+{{ $actionLabel := .actionLabel | default "" }}
```
MykytaPimonovTD: For params in hugo use snake_case. This params is case insensitive. 

Inside file use camel case, for passing use snake case. Here it should looks like:


```suggestion
{{ $actionLabel := .action_label | default "" }}
```

Update everywhere.

site/layouts/_partials/components/result-panel.html
```diff
@@ -0,0 +1,31 @@
...
+{{ $actionUrl := .actionUrl | default "/" }}
```
MykytaPimonovTD: action with params is separate dict.

site/layouts/checkout-completed/single.html
```diff
@@ -0,0 +1,26 @@
+{{ define "main" }}
```
MykytaPimonovTD, line 1: For your information, this is the only type of file where you don't need to include a copyright notice. The line {{.define “main” }} is there because Hugo currently cannot properly process comments in such files.

site/layouts/checkout-completed/single.html
```diff
@@ -0,0 +1,26 @@
...
+                            "title" "Thank you for your purchase"
+                            "titleTag" "h1"
+                            "lines" (slice
+                                "Your payment was completed successfully."
+                                "We will send the order details to your email."
+                            )
+                            "actionLabel" "Back to home"
+                            "actionUrl" "/"
+                        ) }}
+                        {{ .Content }}
```
MykytaPimonovTD: Delete this line. It is no content.

site/layouts/checkout-completed/single.html
```diff
@@ -0,0 +1,26 @@
+{{ define "main" }}
+    {{ partial "components/navbar/navbar.html" . }}
+    {{ partial "components/nav-hero.html" . }}
```
MykytaPimonovTD: Do we need hero here? I do not think so.

site/layouts/checkout/single.html
```diff
@@ -0,0 +1,275 @@
...
+                                </div>
+                                <div class="form-field form-field-full">
+                                    <label class="form-label" for="checkout-phone">Phone number</label>
+                                    <div class="phone-field" data-phone-country-selected="false">
+                                        <select class="form-input form-select phone-country-select"
+                                                id="checkout-phone-country-code"
+                                                name="phone_country_code"
+                                                autocomplete="tel-country-code"
+                                                aria-label="Phone country code">
+                                            <option value="" data-flag="" data-code="" selected>Country</option>
```
MykytaPimonovTD, line 139: See how we did [this](https://teamdev.cn/jxbrowser/#evaluate). I just do not like idea store icons like this.

site/layouts/checkout/single.html
```diff
@@ -0,0 +1,275 @@
...
+                                               id="checkout-phone"
+                                               name="phone_number"
+                                               type="tel"
+                                               autocomplete="tel-national"
+                                               inputmode="tel"
+                                               disabled>
+                                    </div>
+                                </div>
+                                <div class="form-field">
+                                    <label class="form-label" for="checkout-first-name">First name</label>
```
MykytaPimonovTD: We have only one name.

site/assets/js/modules/forms/phone-number.js
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Checks whether the input contains only supported phone-number characters.
+ *
+ * Allowed: digits, parentheses, hyphens, and spaces.
+ *
+ * @param {string} value phone-number value to check
+ * @return {boolean} true when the value contains only allowed characters
+ */
+export function isValidPhoneNumberInput(value) {
+    return /^[0-9\s()-]+$/.test(value);
```
MykytaPimonovTD: Do we really need this? It looks were weird. Phone number it is only numbers.

site/assets/js/modules/paygate/purchases.js
```diff
@@ -0,0 +1,234 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+'use strict';
+
+/**
+ * Paygate product data returned for a checkout order.
```
MykytaPimonovTD: Is it right place for doc? Maybe we should place type doc in file where this type is used? Not sure.

site/assets/js/modules/paygate/purchases.js
```diff
@@ -0,0 +1,234 @@
...
+
+    return {
+        placeOrder(productId) {
+            return postJson(`${baseUrl}/purchases/place-order`, {productId});
+        },
+        calculateCharges(payload) {
+            return postJson(`${baseUrl}/purchases/calculate-charges`, payload);
+        },
+        submitBillingInfo(payload) {
+            return postJson(`${baseUrl}/purchases/submit-billing-info`, payload);
```
MykytaPimonovTD:
```suggestion
            return postJson(`${baseUrl}/purchases/provide-billing-info`, payload);
```

site/assets/js/modules/paygate/purchases.js
```diff
@@ -0,0 +1,234 @@
...
+    };
+}
+
+/**
+ * Removes trailing slashes from the configured Paygate server URL.
+ *
+ * @param {string} url raw configured Paygate server URL
+ * @return {string} server URL without trailing slashes
+ */
+export function normalizeServerUrl(url) {
```
MykytaPimonovTD: I think this is private function. If not, it is very strange.

site/assets/js/modules/paygate/purchases.js
```diff
@@ -0,0 +1,234 @@
...
+    }
+}
+
+/**
+ * Extracts a human-readable error message from a parsed response body.
+ *
+ * @param {*} body parsed response body
+ * @return {string} message text when present, otherwise an empty string
+ */
+function responseMessage(body) {
```
MykytaPimonovTD: `errorMessage`

site/assets/js/pages/checkout/charge-controller.js
```diff
@@ -0,0 +1,305 @@
...
+     */
+    function flush() {
+        clearScheduledRequest();
+        return requestIfReady();
+    }
+
+    /**
+     * Debounces charge recalculation while the user types a VAT ID.
+     */
+    function schedule() {
```
MykytaPimonovTD: Send request on changing focus.

https://github.com/user-attachments/assets/25fe66ee-cd7e-470b-9327-515a13a2c1ed

site/assets/js/pages/checkout/view-controller.js
```diff
@@ -0,0 +1,179 @@
...
+    function isFormHidden() {
+        return dom.$form.prop('hidden');
+    }
+
+    /**
+     * Fills the order summary with product details returned by Paygate.
+     *
+     * @param {Object} product paygate product data for the current order
+     */
+    function hydrateProductSummary(product) {
```
MykytaPimonovTD:
```suggestion
    function fillProductSummary(product) {
```

MykytaPimonovTD (changes requested):
@Vladyslav-Kuksiuk see the first bunch of comments. 12/22.

site/assets/js/modules/forms/phone-number.js
Vladyslav-Kuksiuk: It is for case, when user copied number with those symbols, like (000) 000-0000

site/assets/js/modules/paygate/purchases.js
Vladyslav-Kuksiuk: It is a place where this type is defined, because it is a part of Purchases client API which is defined here.

site/layouts/checkout/single.html
Vladyslav-Kuksiuk: They are merged on the backend, but all the checkout forms I’ve seen use separate fields, so I suggest keeping them separate.

site/assets/js/modules/forms/phone-number.js
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Removes characters that are not accepted by the phone-number field.
+ *
+ * Allowed: digits, parentheses, hyphens, and spaces.
+ *
+ * @param {string} value phone-number value to sanitize
+ * @return {string} sanitized phone-number value
+ */
+export function sanitizePhoneNumberInput(value) {
```
dmytro-kashcheiev, line 37: Is this method used anywhere? Doesn't the above one help us to skip this one?

site/assets/js/modules/paygate/purchases.js
```diff
@@ -0,0 +1,234 @@
...
+ */
+
+/**
+ * Creates a client for Paygate purchase endpoints.
+ *
+ * @param {string} serverUrl base URL of the Paygate API server
+ * @return {PaygatePurchaseClient} paygate purchase endpoint methods
+ */
+export function createPurchaseClient(serverUrl) {
+    const baseUrl = normalizeServerUrl(serverUrl);
```
dmytro-kashcheiev: What is the purpose of this call? Can't the server URL be normalized manually since it's a config value? If for some reason Hugo adds some noise to the string, can we look for the way to remove those noise from the Hugo side?

site/assets/js/modules/paygate/purchases.js
```diff
@@ -0,0 +1,234 @@
...
+ *
+ * @param {Response} response fetch response to parse
+ * @return {Promise<*>} parsed JSON, text, or null when there is no readable body
+ */
+async function readResponseBody(response) {
+    const contentType = response.headers.get('content-type') || '';
+
+    if (response.status === 204) {
+        return null;
+    }
```
dmytro-kashcheiev: This an below response - to - null conversion should be documented in the doc now it's unclear how we handle response. 
Maybe the method above should be updated somehow as well.

site/assets/js/pages/checkout/charge-controller.js
dmytro-kashcheiev, line 1: This file currently unites two big parts of logic: 
1. Request - Reponse to the server.
2. Complicated service that manages request queue. 

Lets divide this on two parts, the simpliest way to do this is imaging that there is, for example, some other input that requires the same `input - request` logic but with totally different endpoints and requests.

site/assets/js/pages/checkout/index.js
```diff
@@ -0,0 +1,254 @@
...
+         *
+         * @param {Object|Error} error request error to inspect
+         * @return {boolean} true when the error represents a server response
+         */
+        function showServerErrorModal(error) {
+            if (error.status >= 500) {
+                return false;
+            }
+
+            view.showErrorModal();
```
dmytro-kashcheiev: Not sure about this. 
But why we do not `view.showErrorModal()`  on 500? Maybe document that not all errors shows modal.

site/assets/js/pages/checkout/index.js
```diff
@@ -0,0 +1,254 @@
...
+                const redirectUrl = response.paymentLink ||
+                    response.redirectUrl ||
+                    response.url ||
+                    response.link;
+
+                if (redirectUrl) {
+                    window.location = redirectUrl;
+                } else {
+                    console.log('Billing info response:', response);
+                }
```
dmytro-kashcheiev: What is the purpose of this code? If after submit there is no redirect maybe something should be done? Not simply write to log.

dmytro-kashcheiev (changes requested):

dmytro-kashcheiev (changes requested):
see comments.

site/layouts/checkout/single.html
Vladyslav-Kuksiuk: As discussed, it is done with library on `teamdev.com`, but here we don't want to introduce it now.

dmytro-kashcheiev (changes requested):
@Vladyslav-Kuksiuk its ok now from the code perspective, but I want to request a few more features from UI-UX: 
1) Since VAT ID is a toxic field, I propose to add an in-progress spinner somewhere in the field that shows that we are actually validating it right now and it's not either aprooved / rejected. 
2) IF VAT passes verification, highlight the field a little green and even maybe small text like: `Provided ID is active and valid`.
3) Lets fix the following behavior: When I have already filled everything and then press `Continue to payment`. If on a new page, I decided I want to double check my address and press the return button. I lose part of the filled data and proceed back.

MykytaPimonovTD:
@Vladyslav-Kuksiuk General comment. Let’s make the product ID a path variable instead of an argument.

The current issue is that a customer could effectively specify any ID, and our payment system would accept it even if it isn’t the expected Spine product.

If we specify products as path variables—like `/checkout/spine-support-50h`—this will require us to add a new content file to each product page, where the actual ID will be specified in the frontmetters. And we decide which product pages to display on the website.

I believe this approach has another advantage: logically, we’ll have a separate page for each product, which makes sense for a static site.

@dmytro-kashcheiev what do you think?

site/assets/scss/modules/_result-panel.scss
```diff
@@ -0,0 +1,165 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+.result-page {
```
JuliaEvseeva, line 27: Please document this component class that it will be used in the 404 pages for different status blocks.

site/assets/scss/pages/_checkout.scss
```diff
@@ -0,0 +1,389 @@
...
+    margin-top: auto;
+  }
+
+  .wrapper,
+  .main,
+  .footer,
+  #header {
+    overflow-x: hidden;
+  }
+}
```
JuliaEvseeva, line 32: These styles are not needed.

Instead keep only the following:
```
.checkout-page {
  .main {
    display: flex;
    flex-direction: column;
  }
}
```

site/assets/scss/pages/_checkout.scss
```diff
@@ -0,0 +1,389 @@
...
+
+.checkout {
+  flex: 1 0 auto;
+  display: flex;
+  flex-direction: column;
+  background:
+    radial-gradient(circle at top left, rgba($main-brand-color, .10), transparent 36%),
+    linear-gradient(180deg, #f5faff 0%, var(--body-bg-color) 220px);
+  overflow-x: hidden;
```
JuliaEvseeva: Avoid using this property. If the layout breaks on mobile and creates horizontal scrolling, adjust the layout rather than cropping it.

site/assets/scss/pages/_checkout.scss
```diff
@@ -0,0 +1,389 @@
...
+
+  .article-container {
+    height: 100%;
+  }
+}
+
+.checkout-summary {
+  max-width: 100%;
+  min-width: 0;
+  margin-bottom: 8px;
```
JuliaEvseeva: The following styles are not needed.

site/assets/scss/pages/_checkout.scss
```diff
@@ -0,0 +1,389 @@
...
+  .checkout-summary-loading {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+    width: 100%;
+    max-width: 420px;
+    margin: 0 auto;
+    color: $gray-500;
+    font-size: 17px;
```
JuliaEvseeva: Let's use the default `16px` font-size here.

site/assets/scss/pages/_checkout.scss
```diff
@@ -0,0 +1,389 @@
...
+      }
+    }
+  }
+
+  .checkout-summary-row {
+    display: grid;
+    grid-template-columns: minmax(0, 1fr) auto;
+    align-items: end;
+    column-gap: 20px;
+    padding: 12px 0 16px;
```
JuliaEvseeva: Let's use the same top and bottom paddings to align the content inside the row.

```suggestion
    padding: 12px 0;
```

site/assets/scss/pages/_checkout.scss
```diff
@@ -0,0 +1,389 @@
...
+    .checkout-summary-value {
+      font-size: 20px;
+      font-weight: 800;
+      letter-spacing: -.02em;
+    }
+  }
+
+  .checkout-summary-product {
+    max-width: 100%;
+    padding: 2px 0 8px;
```
JuliaEvseeva:
```suggestion
    padding-bottom: 8px;
```

site/assets/scss/pages/_checkout.scss
```diff
@@ -0,0 +1,389 @@
...
+    line-height: 1.12;
+    overflow-wrap: anywhere;
+
+    @include breakpoint(md-phone) {
+      font-size: 32px;
+    }
+  }
+
+  .checkout-summary-product-description {
+    margin: -2px 0 10px;
```
JuliaEvseeva:
```suggestion
    margin-bottom: 40px;
```

Vladyslav-Kuksiuk:
> @Vladyslav-Kuksiuk General comment. Let’s make the product ID a path variable instead of an argument.
> 
> The current issue is that a customer could effectively specify any ID, and our payment system would accept it even if it isn’t the expected Spine product.
> 
> If we specify products as path variables—like `/checkout/spine-support-50h`—this will require us to add a new content file to each product page, where the actual ID will be specified in the frontmetters. And we decide which product pages to display on the website.
> 
> I believe this approach has another advantage: logically, we’ll have a separate page for each product, which makes sense for a static site.
> 
> @dmytro-kashcheiev what do you think?

1. If we don’t have such a product, the server will respond with a 404, so the user won’t be able to specify any product. (When we introduce JxBrowser products, we may use different endpoints.)
2. I’m not sure we always want to represent a new product as a separate page. For example, if a customer wants a custom Spine support pack (this wasn’t part of the previous discussion, but I believe that if demand grows, these support packs will become configurable), creating a separate page for each product would be cumbersome and wouldn’t provide any real benefit.
3. In one of our discussions, we mentioned that we might want to include a company as part of the `place-order` request. In that case, it may look inconsistent if the product and company are specified differently.

site/assets/scss/pages/_checkout.scss
```diff
@@ -0,0 +1,389 @@
...
+    .checkout-summary-row-total .checkout-summary-label,
+    .checkout-summary-value-total {
+      font-size: 18px;
+    }
+  }
+}
+
+.checkout .form-section {
+  margin-top: 12px;
+  padding-top: 4px;
```
JuliaEvseeva: Here you override styles from `site/assets/scss/modules/_forms.scss`. Why are they defined there if they’re only used for this form?

Please keep only `margin-top: 8px;` here and remove everything else in the `forms.scss` as well for the class `form-section`.

site/assets/scss/modules/_forms.scss
```diff
@@ -0,0 +1,339 @@
...
+}
+
+@keyframes vat-field-spin {
+  to {
+    transform: translateY(-50%) rotate(360deg);
+  }
+}
+
+.form-actions {
+  margin-top: 20px;
```
JuliaEvseeva:
```suggestion
  margin: 40px 0 8px;
```

site/assets/scss/modules/_forms.scss
```diff
@@ -0,0 +1,339 @@
...
+  margin-top: 24px;
+  padding-top: 20px;
+  border-top: 1px solid rgba(black, .08);
+}
+
+.form-title {
+  margin-bottom: 16px;
+  padding-top: 0;
+  color: $black;
+  font-size: 24px;
```
JuliaEvseeva:
```suggestion
  font-size: 24px !important;
```

site/assets/scss/modules/_forms.scss
```diff
@@ -0,0 +1,339 @@
...
+  &:hover,
+  &:focus {
+    text-decoration: none;
+  }
+}
+
+@include breakpoint(lg-phone) {
+  .form-title {
+    font-size: 24px;
+  }
```
JuliaEvseeva: This override isn't needed because the value is the same as on the desktop.

MykytaPimonovTD:
> > @Vladyslav-Kuksiuk General comment. Let’s make the product ID a path variable instead of an argument.
> > The current issue is that a customer could effectively specify any ID, and our payment system would accept it even if it isn’t the expected Spine product.
> > If we specify products as path variables—like `/checkout/spine-support-50h`—this will require us to add a new content file to each product page, where the actual ID will be specified in the frontmetters. And we decide which product pages to display on the website.
> > I believe this approach has another advantage: logically, we’ll have a separate page for each product, which makes sense for a static site.
> > @dmytro-kashcheiev what do you think?
> 
> 1. If we don’t have such a product, the server will respond with a 404, so the user won’t be able to specify any product. (When we introduce JxBrowser products, we may use different endpoints.)
> 2. I’m not sure we always want to represent a new product as a separate page. For example, if a customer wants a custom Spine support pack (this wasn’t part of the previous discussion, but I believe that if demand grows, these support packs will become configurable), creating a separate page for each product would be cumbersome and wouldn’t provide any real benefit.
> 3. In one of our discussions, we mentioned that we might want to include a company as part of the `place-order` request. In that case, it may look inconsistent if the product and company are specified differently.

@Vladyslav-Kuksiuk OK. I agree.

site/layouts/_partials/components/result-panel.html
```diff
@@ -0,0 +1,65 @@
...
+        {{ $actionLabel := .label | default .text | default "" }}
+        {{ $actionUrl := .url | default site.Home.RelPermalink }}
+        {{ if $actionLabel }}
+        <form class="form-actions" action="{{ $actionUrl }}" method="get">
+            <button class="btn btn-bordered-blue form-submit-button"
+                    type="submit">
+                {{ $actionLabel }}
+            </button>
+        </form>
+        {{ end }}
```
JuliaEvseeva: Please align the code.

I don’t think this is actually a "form actions". It doesn’t submit anything and probably never will.

site/layouts/checkout-completed/single.html
```diff
@@ -0,0 +1,29 @@
...
+                                "label" "Back to home"
+                                "url" site.Home.RelPermalink
+                            )
+                        ) }}
+                    </div>
+                </div>
+            </div>
+        </div>
+    </section>
+    {{ partial "components/go-top-button.html" . }}
```
JuliaEvseeva: Let’s not add a "Go to top" button. It only makes sense for long pages with a lot of content.

site/layouts/404.html
```diff
@@ -0,0 +1,30 @@
...
+                                "label" "Back to home"
+                                "url" site.Home.RelPermalink
+                            )
+                        ) }}
+                    </div>
+                </div>
+            </div>
+        </div>
+    </section>
+    {{ partial "components/go-top-button.html" . }}
```
JuliaEvseeva: Please delete the import of the "Go to top" button component.

JuliaEvseeva (changes requested):

site/assets/js/modules/forms/phone-number.js
```diff
@@ -0,0 +1,66 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+'use strict';
+
+/**
+ * Removes characters that are not accepted by the phone-number field.
+ *
+ * Allowed: digits, parentheses, hyphens, and spaces.
```
MykytaPimonovTD:
```suggestion
 * <p>Allowed: digits, parentheses, hyphens, and spaces.
```

site/assets/js/pages/checkout/charge-controller.js
```diff
@@ -0,0 +1,237 @@
...
+
+import {createDelayedRequestController} from 'js/pages/checkout/delayed-request-controller';
+import {fieldValidationState} from 'js/pages/checkout/form-controller';
+
+/**
+ * Delay before sending the 'calculate-charges' request when VAT ID was changed.
+ *
+ * @type {number}
+ */
+const vatIdInputDelay = 1000;
```
MykytaPimonovTD, line 43: 1000 of what. Mention that it is millis in docs.

site/layouts/_partials/components/result-panel.html
```diff
@@ -0,0 +1,65 @@
...
+  ~ OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+  ~ SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+  ~ LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+  ~ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+  ~ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+  ~ (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+  ~ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+  -->
+
+<!-- Result panel is a reusable centered status block -->
```
MykytaPimonovTD:
```suggestion
<!-- Result panel is a reusable centered status block. -->
```

dmytro-kashcheiev (approved):
@Vladyslav-Kuksiuk LGTM from my side.

MykytaPimonovTD (approved):
@Vladyslav-Kuksiuk LGTM.

JuliaEvseeva (approved):

PR #543, opened 2026-05-22
Include validation docs

alexander-yevsyukov:
This PR updates references to Hugo modules of Documentation and Validation documentation so that Validation appears in the navigation bar and all.


alexander-yevsyukov:
@JuliaEvseeva, I'll need to add link verification to the Validation Docs project and fix them there. Then, I'll return to this PR.

alexander-yevsyukov:
We don't need to add Validation docs as a submodule because it is added to the Documentation project instead.

PR #548, opened 2026-08-13
Update checkout page

Vladyslav-Kuksiuk:
This PR updates the checkout page and checkout-completed page according to the latest changes from the Company-Site versions of those pages.

### Checkout page

image
image

### Order not found
image

### Transaction in progress

image

### Payment completed

image

### Payment failed

image

Resolves [this issue](https://github.com/TeamDev-Ltd/Paygate/issues/125).


site/assets/scss/modules/_forms.scss
JuliaEvseeva: Let's use the project colors here:

```suggestion
$form-error-color: $red;
$form-success-color: $green;
```

JuliaEvseeva: There is no need to colorize the label on the error state.

JuliaEvseeva:
```suggestion
  color: $form-error-color;
```

JuliaEvseeva:
```suggestion
    border-color: $form-error-color;
```

site/assets/scss/modules/_result-panel.scss
JuliaEvseeva, line 26: Please add this part at the top of the file:

```
.error-pages {
  .main {
    display: flex;
    flex-direction: column;
  }
}
```

site/assets/scss/pages/_checkout.scss
```diff
@@ -24,168 +24,136 @@
...
 .checkout {
   flex: 1 0 auto;
   display: flex;
   flex-direction: column;
-  background: radial-gradient(circle at top left, rgba($main-brand-color, .10), transparent 36%),
-  linear-gradient(180deg, #f5faff 0%, var(--body-bg-color) 220px);
+  width: 100%;
+  min-height: 0;
+  overflow-x: hidden;
+  background: $checkout-bg-color;
```
JuliaEvseeva: Let's return the background gradient that was in the previous implementation.

site/assets/scss/pages/_checkout.scss
```diff
@@ -24,168 +24,136 @@
...
+  width: 100%;
 
-      .loader {
-        width: 34px;
-        height: 34px;
-      }
+  .checkout-summary-product {
+    max-width: 100%;
+    padding-bottom: 10px;
+    color: $black;
```
JuliaEvseeva: Please check everywhere that the `$text-color` variable is used for text colors instead of `$black`:

```suggestion
    color: $text-color;
```

site/assets/scss/pages/_checkout.scss
```diff
@@ -24,168 +24,136 @@
...
-    .checkout-summary-loading {
-      font-size: 16px;
+.checkout-summary {
+  width: 100%;
 
-      .loader {
-        width: 34px;
-        height: 34px;
-      }
+  .checkout-summary-product {
```
JuliaEvseeva, line 146: Let's revert the page title font styles to match the previous implementation.

site/assets/scss/pages/_checkout.scss
```diff
@@ -194,171 +162,282 @@
...
+  }
 
-    .checkout-summary-row {
-      column-gap: 12px;
-      padding: 10px 0 12px;
-    }
+  .form-submit-button.disabled,
+  .form-submit-button:disabled {
+    cursor: not-allowed;
+    opacity: .55;
```
JuliaEvseeva:
```suggestion
    opacity: .4;
```

site/assets/scss/modules/_forms.scss
```diff
@@ -106,6 +106,19 @@ $form-success-color: #2e7d32;
   padding-right: 44px;
 }
 
+.form-field-vat {
```
JuliaEvseeva, line 105: Please also try to keep the same behavior as in JxBrowser.
1. The input shouldn't show an error while the user is typing.
2. Displaying the error message shouldn't shift or move the content below it.

Should be:

https://github.com/user-attachments/assets/1a6d360e-3527-4ed9-b134-4aa3f489c2b9

Now:

https://github.com/user-attachments/assets/19e861b7-c23b-4550-bee6-438e3398d367

JuliaEvseeva (changes requested):
@Vladyslav-Kuksiuk, please see minor UI comments.

site/layouts/_partials/components/result-panel.html
```diff
@@ -44,19 +45,28 @@ <h1 class="result-panel-title">{{ $title }}</h1>
...
+        {{ $actionLabel := .label | default .text | default "" }}
+        {{ $actionUrl := .url | default site.Home.RelPermalink }}
+        {{ if $actionLabel }}
+        <a class="btn {{ if .primary }}btn-primary-blue{{ else }}btn-bordered-blue{{ end }} result-panel-button"
+           href="{{ $actionUrl }}"
+           {{ if .back_to_checkout }}data-back-to-checkout hidden{{ end }}>
+            {{ $actionLabel }}
+        </a>
+        {{ end }}
+        {{ end }}
```
JuliaEvseeva: Let's indent the code inside the range.

site/layouts/checkout-completed/single.html
```diff
@@ -1,24 +1,126 @@
 {{ define "main" }}
     {{ partial "components/navbar/navbar.html" . }}
-    <section class="checkout checkout-completed">
-        <div class="container-fluid content-holder content-with-fixed-header">
-            <div class="row justify-content-center">
-                <div class="col-lg-7">
-                    <div class="article-container article-text">
+    {{ $email := hugo.Data.emails.sales_email }}
+    {{ $contactLine := printf "If you have any questions, email us at <a href=\"mailto:%s\">%s</a>." $email $email | safeHTML }}
+    {{ $missingOrderContactLine := printf "Please check the address or contact <a href=\"mailto:%s\">%s</a>." $email $email | safeHTML }}
```
JuliaEvseeva: Let's improve this part to use the `cloakemail`:

```
{{ $salesEmail := hugo.Data.emails.sales_email }}
{{ $salesEmailLink := (printf "{{< cloakemail address=\"%s\" >}}" $salesEmail) }}

{{ $contactLine := printf "If you have any questions, email us at %s." $salesEmailLink }}
{{ $missingOrderContactLine := printf "Please check the address or contact %s." $salesEmailLink }}
```

In the file `result-panel.html` we need to fix the code alignment and apply the `markdownify` function to all the relevant markdown fields.

site/layouts/checkout/single.html
```diff
@@ -1,280 +1,257 @@
 {{ define "main" }}
     {{ partial "components/navbar/navbar.html" . }}
+    {{ $email := hugo.Data.emails.sales_email }}
+    {{ $emailLink := printf "<a href=\"mailto:%s\">%s</a>" $email $email | safeHTML }}
+    {{ $missingOrderContactLine := printf "Please check the address or contact <a href=\"mailto:%s\">%s</a>." $email $email | safeHTML }}
```
JuliaEvseeva: Same as above.

site/assets/scss/pages/_checkout.scss
```diff
@@ -24,206 +24,129 @@
...
-  }
+.checkout-summary {
+  width: 100%;
 
   .checkout-summary-product {
     max-width: 100%;
     padding-bottom: 8px;
-    color: rgba($black, .4);
+    color: $text-color;
     font-size: 42px;
```
JuliaEvseeva:
```suggestion
    color: rgba(black, .4);
    font-size: $font-size--xxxl;
```

JuliaEvseeva (approved):

site/assets/js/pages/checkout/form-controller.js
```diff
@@ -169,14 +306,15 @@ export function createCheckoutFormController({dom}) {
         const formData = Object.fromEntries(new FormData(dom.form).entries());
         const field = name => (formData[name] || '').trim();
         const companyName = field('company');
-        const vatId = field('vat_id');
+        const vatId = getVatId();
         const fullName = [field('first_name'), field('last_name')]
```
Oleg-Melnik, line 311: `billingInfo.name` can now be submitted empty. `first_name`/`last_name` were always optional, and this PR also dropped `required` from `#checkout-company`, so a buyer who fills only country / email / address line 1 / city passes `validateRequiredFields()` and `fullName` resolves to `''`.

The `company` object four lines below has the mirror problem: a VAT ID entered without a company name produces `company: {vatId: 'X'}` with no `name`, which does not match the `BillingCompany` shape declared in `purchases.js`.

Both cases only surface as the generic "Something went wrong" modal, after the user has filled in the entire form.

site/assets/js/pages/checkout/index.js
```diff
@@ -222,24 +237,74 @@ $(
...
+        function scheduleCurrentChargeCalculation() {
+            window.setTimeout(requestCurrentCharges, 0);
+        }
+
+        /** Restores custom country widgets after native browser form restoration settles. */
+        function scheduleCheckoutRestoration() {
+            window.setTimeout(() => {
+                const restoredState = formController.getBrowserRestoredCountryState();
+
+                phoneCountryManuallySelected = Boolean(
```
Oleg-Melnik: This derivation misfires for the common case where the user never touched the phone country.

The phone field is initialised to `us` (`initPhoneNumberField()`), and `syncPhoneCountryState()` writes that into `#checkout-phone-country`. So a visitor who opens checkout, types their email, navigates away and presses Back restores `phoneCountryCode: 'US'` with `billingCountryCode: ''` — and `'US' !== ''` sets `phoneCountryManuallySelected = true`.

From then on `applyPhoneCountryFromBillingCountry()` is a no-op, so selecting "Germany" as the billing country leaves the phone prefix at +1.

On a bfcache restore this is doubly wrong: the closure variable survived the navigation and already held the correct `false`, and this line overwrites it.

site/layouts/checkout-completed/single.html
```diff
@@ -1,24 +1,128 @@
...
+                                "The checkout order could not be found."
+                            )
+                            "additional_line" $missingOrderContactLine
+                            "actions" (slice (dict
+                                "label" "Back to home"
+                                "url" site.Home.RelPermalink
+                            ))
+                        ) }}
+                    </div>
+                    <div id="payment-status-unknown">
```
Oleg-Melnik, line 114: `#payment-status-unknown` is the only view rendered without `hidden`, which makes "Thank you! We will send the order details to your email after the payment is confirmed." the page's default state.

`completed.js` `init()` returns early — without ever calling `showView()` — when the `orderId` query parameter is missing, when `params.payment.paygateurl` is unset, and when `createPurchaseClient()` throws. In all three cases (and with JS disabled) a visitor who has no order at all is told their payment is being confirmed.

`#payment-order-not-found` looks like the right default for the missing-`orderId` path.

site/assets/js/pages/checkout/completed.js
```diff
@@ -0,0 +1,214 @@
...
+        }
+        pollingTimerId = window.setTimeout(
+            pollPaymentStatus,
+            Math.min(delay, remainingDurationMs)
+        );
+    }
+
+    /** Stops polling when its maximum duration is reached. */
+    function reachPollingDeadline() {
+        finishWithView(viewIds.UNKNOWN);
```
Oleg-Melnik, line 167: Reaching the 15-minute deadline always lands on `UNKNOWN` ("Thank you!"), including the case where every single poll failed and `STATUS_ERROR` is currently on screen.

After 15 minutes of 5xx/network errors the buyer is told the order details are on their way, even though the site never received a single successful status response. Tracking whether any poll succeeded and keeping `STATUS_ERROR` when none did would avoid asserting an outcome that was never observed.

site/layouts/_partials/getting-help/services.html
```diff
@@ -41,8 +41,12 @@ <h1 class="section-title light">{{ $data.title | markdownify }}</h1>
             <div class="support-card compare-support">
                 {{ $comparableServices := $data.comparable_services }}
                 {{ range $comparableServices.header_cols }}
+                    {{ $productId := .paygate_product_id | default "" }}
+                    {{ with site.Params.payment.standardsupportproductid }}
+                        {{ $productId = . }}
+                    {{ end }}
```
Oleg-Melnik: The override sits inside the `range`, so it replaces the product ID of *every* column rather than just the standard-support one.

`services.json` has two `header_cols`: "Standard Support Pack" (with `paygate_product_id`) and "Custom Support" (deliberately without one). With `standardSupportProductId` configured, the second column also gets `data-paygate-product-id`. It happens to be harmless today only because `pricing.js` reads the first match of `$('[data-paygate-product-id]')`.

Scoping the override to columns that already declare a product ID keeps the intent:

```suggestion
                    {{ if $productId }}
                        {{ with site.Params.payment.standardsupportproductid }}
                            {{ $productId = . }}
                        {{ end }}
                    {{ end }}
```

site/assets/js/pages/checkout/view-controller.js
```diff
@@ -110,46 +118,62 @@ export function createCheckoutView(dom) {
...
+        dom.$loading.prop('hidden', true);
         dom.$summary.prop('hidden', true);
         dom.$form.prop('hidden', true);
+        dom.$missingOrder.prop('hidden', true);
         dom.$notFound.prop('hidden', true);
         dom.$summaryError.prop('hidden', false);
     }
 
+    /** Shows the resolved order summary and billing form. */
+    function showCheckoutView() {
```
Oleg-Melnik, line 157: `showSummaryLoading`, `showSummaryError`, `showCheckoutView`, `showMissingOrderView` and `showNotFoundView` are five copies of the same six-line block — toggle `$loading` / `$summary` / `$form` / `$missingOrder` / `$notFound` / `$summaryError`, differing only in which one stays visible plus the `setResultPageMode` flag.

`completed.js` already has the generic form of exactly this (`showView(activeViewId)` iterating the view IDs). A small map from panel name to `{element, isResultPage}` would drop ~30 lines here and make it impossible to forget a panel when a sixth one is added.

site/tests/checkout-configuration.test.mjs
```diff
@@ -0,0 +1,585 @@
...
+        '../assets/js/pages/checkout/charge-controller.js'
+    ).replace(/^import .*;\n/gm, '');
+    const dependencies = `
+        const fieldValidationState = {idle: 'idle', loading: 'loading', success: 'success'};
+        const buildChargeRequest = (orderId, buyerCountryCode, vatId) => {
+            if (!orderId || !buyerCountryCode) return null;
+            return vatId
+                ? {orderId, buyerCountryCode, vatId}
+                : {orderId, buyerCountryCode};
+        };
```
Oleg-Melnik: `buildChargeRequest` is re-implemented here instead of being imported, and the copy is byte-for-byte the current `charge-request.js` body.

The real module is already loaded at the top of this file via `importSource('../assets/js/pages/checkout/charge-request.js')`, so the harness can inject that export. As written, a behaviour change in `charge-request.js` cannot fail any charge-controller test, and the two copies can drift silently.

.github/workflows/check-links.yml
```diff
@@ -108,6 +108,11 @@ jobs:
         working-directory: ${{ steps.docs.outputs.work_dir }}
         run: npm ci
 
+      - name: Run checkout tests
+        if: steps.docs.outputs.present == 'true'
+        working-directory: ${{ steps.docs.outputs.work_dir }}
+        run: npm test
```
Oleg-Melnik: The JS unit suite is wired into the *link-check* workflow and gated on `steps.docs.outputs.present == 'true'`, so it only runs when a docs preview is being built, and a unit-test failure is reported to contributors as a link-check failure.

These tests do not depend on the rendered site at all. A dedicated job (or an existing build workflow) would run them on every PR and report under their own name.

site/assets/js/pages/checkout/countries.js
```diff
@@ -0,0 +1,131 @@
...
+        ? new Intl.DisplayNames(['en'], {type: 'region'})
+        : null;
+    const countries = countryCodes.map(code => ({
+        code,
+        name: displayNames ? displayNames.of(code) : code
+    })).sort((first, second) => first.name.localeCompare(second.name));
+
+    countries.forEach(country => {
+        select.add(new Option(country.name, country.code));
+    });
```
Oleg-Melnik, line 67: This appends ~250 options one at a time, each mutating the live `<select>` during page load and immediately before Select2 re-reads it. Building into a `DocumentFragment` and appending once makes it a single mutation:

```suggestion
    const options = document.createDocumentFragment();
    countries.forEach(country => {
        options.append(new Option(country.name, country.code));
    });
    select.append(options);
}
```

site/assets/scss/pages/_checkout.scss
```diff
@@ -236,129 +159,286 @@
...
+
+  .form-input,
+  .form-select {
+    width: 100%;
+    min-height: 48px;
+  }
+
+  .form-select {
+    appearance: auto;
+  }
```
Oleg-Melnik: `.form-select` is dead under `.checkout .form-section` after this PR: `#checkout-country` changed from `class="form-input form-select"` to `class="form-input country-select"`, and the phone country `<select>` was removed entirely. Nothing in the checkout markup carries the class any more.

This whole `appearance: auto` rule can go, and `.form-select` can be dropped from the `.form-input, .form-select` selector on line 247.

site/assets/js/pages/checkout/form-controller.js
```diff
@@ -117,6 +169,15 @@ export function createCheckoutFormController({dom}) {
             return true;
         }
 
+        if (field.disabled || field.closest('[hidden]')) {
+            setFieldError(field, '');
+            return true;
+        }
+
+        if (field.type === 'tel') {
```
Oleg-Melnik: This returns `true` for any `type="tel"` field *before* the `field.required` check, so a required phone input would silently pass `validateRequiredFields()`.

Today it is unreachable — `#checkout-phone` is not `required`, and `bindEvents()` routes the phone field to its own `blur -> validatePhoneNumber` handler rather than here. It reads like a guard but is dead code that would misbehave the moment it becomes live.

site/assets/js/pages/checkout/index.js
```diff
@@ -222,24 +237,74 @@ $(
...
+            }, 0);
+        }
+
+        /** Clears browser-restored custom checkout fields after an explicit reload. */
+        function scheduleCheckoutReset() {
+            window.setTimeout(() => {
+                phoneCountryManuallySelected = false;
+                formController.restoreCountryState({
+                    billingCountryCode: '',
+                    phoneCountryCode: 'US'
```
Oleg-Melnik: `'US'` duplicates the default phone country from `form-controller.js` (`initialCountry: normalizeCountryCode(dom.$phoneCountry.val()) || 'us'`), in a different case.

Exporting a single default from the form controller would keep the reload reset and the initial render from drifting apart.

Oleg-Melnik (approved):
@Vladyslav-Kuksiuk LGTM with comments to address.

site/assets/js/pages/checkout/form-controller.js
Vladyslav-Kuksiuk: It's an expected behavior.

site/layouts/checkout-completed/single.html
Vladyslav-Kuksiuk: It is an intended behavior for backward compatibility.

site/assets/js/pages/checkout/completed.js
Vladyslav-Kuksiuk: It is an intended behavior.

